PR #120, opened 2026-02-14
ci: gate e2e to main/deploy and add pre-push full checks

Keith-CY:
## Summary
- Add Merge Queue support to CI via `merge_group` trigger so checks can run for queued merge candidates.
- Gate CI e2e job to `main` branch pushes only.
- Require an e2e job before release build/deploy pipeline.
- Add `scripts/run-all-tests.sh` and a pre-push hook to run daemon + gateway checks/tests + e2e entry locally before pushing.
- Add docs for merge queue setup and local pre-push usage.

## Notes
- e2e execution is centralized in `./scripts/run-e2e-tests.sh` so future e2e command can be wired in one place.
- Non-code changes from working tree were intentionally excluded from this PR scope.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
   e2e-deploy:
     name: End-to-End Tests (Deployment)
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/'))
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -109,7 +109,7 @@ jobs:
   release:
     needs: build-packages
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/main'
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/'))
     steps:
       - name: Download packages
         uses: actions/download-artifact@v4


dev01lay2 (commented):
Good additions — pre-push hooks, paths-ignore, merge queue docs, and e2e CI gating. One blocking concern with the e2e script (see inline).

scripts/run-e2e-tests.sh
```diff
@@ -0,0 +1,17 @@
...
+
+repo_root="$(git rev-parse --show-toplevel)"
+
+cd "$repo_root"
+
+echo "Running end-to-end test suite..."
+
+if ! command -v carrier >/dev/null 2>&1; then
+  echo "Error: `carrier` CLI is not available in PATH."
+  echo "Expected e2e command from repository plan: `carrier test e2e --report test-results/`"
```
dev01lay2: BS: The backticks around `carrier` in these echo lines are shell command substitution, not literal formatting. Bash will try to execute `carrier` as a subcommand inside the double-quoted string. Use single quotes or escape them:
```bash
echo "Error: 'carrier' CLI is not available in PATH."
```
(Same issue on lines 13–14.)

.github/workflows/release.yml
```diff
@@ -10,7 +10,29 @@ permissions:
   contents: write
 
 jobs:
+  e2e-deploy:
+    name: End-to-End Tests (Deployment)
```
dev01lay2: BS: `release.yml` triggers on `push: tags: ["v*"]`, so `github.ref` will be `refs/tags/v*`, never `refs/heads/main`. This condition always evaluates false → `e2e-deploy` is always skipped → `build-packages` (which `needs: e2e-deploy`) is also skipped → **releases will stop working**.

Either remove the `if` guard (so e2e runs on every release tag push), or change the condition to match tag refs.

scripts/run-all-tests.sh
```diff
@@ -0,0 +1,22 @@
...
+set -euo pipefail
+
+repo_root="$(git rev-parse --show-toplevel)"
+
+printf '\n=== Daemon tests ===\n'
+(
+  cd "$repo_root/daemon"
+  go test ./...
+  go test ./internal/manifest -run TestLoadFileAcceptsCatalogManifest -count=1
+)
```
dev01lay2: NBS: This specific test is already covered by `go test ./...` on line 10. The explicit `-run` invocation is redundant and can be removed.

PR #265, opened 2026-02-14
test(gateway): guard against duplicate cleanup intervals in DownloadTokenStore

### LINKED ISSUE #235: test(gateway): guard against duplicate cleanup intervals in DownloadTokenStore
## Scope
Add a focused gateway test to verify `startPeriodicCleanup()` can be called multiple times without creating duplicate active timers (regression guard for timer lifecycle).

## Acceptance Criteria
- New test added under `gateway/src/downloads/token_store.test.ts` (or nearby test file).
- Test demonstrates that repeated `startPeriodicCleanup()` calls do not leave multiple active intervals.
- Test is deterministic and does not rely on flaky timing windows.
- `bun test` passes locally and in CI.

dev01lay2:
Closes #235

### DIFF
--- a/gateway/src/downloads/token_store.test.ts
+++ b/gateway/src/downloads/token_store.test.ts
@@ -117,6 +117,52 @@ describe("DownloadTokenStore", () => {
     expect(store.size).toBe(0);
   });
 
+  test("repeated startPeriodicCleanup() does not create duplicate intervals", async () => {
+    let time = new Date("2026-01-01T00:00:00Z");
+    const store = new DownloadTokenStore(() => time);
+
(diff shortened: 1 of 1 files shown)


Keith-CY (commented):
NBS: In `startPeriodicCleanup() replaces previous interval without leak`, the variable `cleanupCount` is currently unused; remove it (or use it to assert cleanup frequency) to avoid a misleading test and keep intent clear.

Akane-CN (approved):
Good regression coverage addition for timer lifecycle behavior.

The new tests cover repeated `startPeriodicCleanup()` calls and guard against interval duplication/leak regressions.

All required checks are green on this head commit.

PR #430, opened 2026-02-14
feat(daemon): HTTP API server

### LINKED ISSUE #286: feat(daemon): HTTP API server for lifecycle commands
## Problem
Daemon `main.go` boots and prints catalog but has no HTTP server — gateway cannot send commands to it.

## Solution
1. Add HTTP server in `daemon/cmd/agentd/main.go` listening on configurable port (default 7331)
2. REST endpoints mapping to lifecycle service:
   - `POST /api/agents` → ListAgents
   - `POST /api/install` → Install
   - `POST /api/start` → Start
   - `POST /api/stop` → Stop
   - `GET /api/status/:agentId` → Status
   - `GET /api/logs/:agentId` → Logs
   - `POST /api/upgrade` → Upgrade
   - `POST /api/diagnose` → Diagnose + consent flow
3. JSON request/response matching `docs/command-contract.md`
4. Integrate with existing health endpoints (/healthz, /readyz)

## Acceptance Criteria
- `curl http://localhost:7331/api/agents` returns agent list
- All lifecycle commands callable via HTTP
- Graceful shutdown stops HTTP server too

## L2 Tasks
- [ ] #302 Define lifecycle HTTP schema and DTO mapping
- [ ] #303 Implement lifecycle handlers, shutdown wiring, and tests

Dependency: #302 -> #303

dev01lay2:
Closes #286

### DIFF
--- a/daemon/cmd/agentd/main.go
+++ b/daemon/cmd/agentd/main.go
@@ -304,6 +304,9 @@ func validateAgentID(id string) error {
 	if strings.Contains(id, "/") || strings.Contains(id, "\\") {
 		return fmt.Errorf("agent ID must not contain path separators")
 	}
+	if strings.Contains(id, "..") {
+		return fmt.Errorf("agent ID must not contain parent-directory tokens")
+	}
 	if !agentIDPattern.MatchString(id) {
 		return fmt.Errorf("agent ID contains invalid characters")
 	}
--- a/daemon/cmd/agentd/main_test.go
+++ b/daemon/cmd/agentd/main_test.go
@@ -146,3 +146,27 @@ func TestAPIMethodNotAllowed(t *testing.T) {
 		t.Fatalf("expected 405, got %d", rec.Code)
 	}
 }
+
+func TestValidateAgentID(t *testing.T) {
+	tests := []struct {
+		id      string
+		wantErr bool
+	}{
+		{"my-agent", false},
+		{"agent_1.0", false},
+		{"a", false},
+		{"", true},
+		{"../etc/passwd", true},
+		{"foo/bar", true},
+		{"foo\\bar", true},
+		{"a..b", true},
+		{".hidden", true},
+		{"-start", true},
+	}
+	for _, tt := range tests {
+		err := validateAgentID(tt.id)
+		if (err != nil) != tt.wantErr {
+			t.Errorf("validateAgentID(%q) err=%v, wantErr=%v", tt.id, err, tt.wantErr)
+		}
+	}
+}


Akane-CN (commented):
I’m holding approval for this head commit because required checks are not green yet.

- Daemon Lint: failing

Please fix the failing check and re-run CI; I’ll re-review promptly after it turns green.

daemon/cmd/agentd/main.go
```diff
@@ -53,19 +57,235 @@ func main() {
...
+	mux.HandleFunc("/api/agents", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			writeJSONError(w, http.StatusMethodNotAllowed, "method not allowed")
+			return
+		}
+		agents := svc.ListAgents()
+		writeJSON(w, http.StatusOK, agents)
+	})
+
+	mux.HandleFunc("/api/install", func(w http.ResponseWriter, r *http.Request) {
```
Keith-CY: BS: `/api/install` is unauthenticated. With configurable bind host, this can become remote install/start/stop control. Please add an auth gate (e.g., bearer token / mTLS / strict loopback enforcement) before exposing mutating endpoints.

daemon/cmd/agentd/main.go
```diff
@@ -53,19 +57,235 @@ func main() {
 		fmt.Printf("- %s (%s): %s\n", entry.Name, entry.ID, entry.Status)
 	}
 
+	// Build HTTP server
+	mux := buildHTTPMux(svc)
+	addr := fmt.Sprintf("%s:%d", cfg.Server.Host, cfg.Server.Port)
```
Keith-CY: BS: Server bind address is fully configurable but there is no transport/auth hardening. Please enforce loopback-only by default with explicit secure opt-in, or require authenticated access before accepting control-plane requests.

dev01lay2:
Addressed the security review: bearer token auth middleware is now applied to all `/api/*` routes. The token is read from config (`api_token`) or env `CARRIER_SERVER_API_TOKEN`. Non-loopback binding is refused without a token. Health checks (`/healthz`, `/readyz`) are exempt. All tests pass. Please re-review.

daemon/cmd/agentd/main.go
```diff
@@ -53,19 +59,247 @@ func main() {
...
+
+	mux.HandleFunc("/api/start", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			writeJSONError(w, http.StatusMethodNotAllowed, "method not allowed")
+			return
+		}
+		var body agentIDBody
+		if !decodeBody(w, r, &body) {
+			return
+		}
```
Keith-CY: BS: `agentID` is taken directly from the URL path without URL-decoding and validation. Please `url.PathUnescape` and reject path separators/invalid IDs before passing into lifecycle methods to avoid ambiguous routing and unexpected agent lookup behavior.

dev01lay2:
Addressed all review comments in ecb7e3a:

1. **Auth gate for mutating endpoints**: Bearer token auth middleware was already in place from ca2148d. When no API token is configured, the server now **forces** bind address to `127.0.0.1` (not just warns), ensuring unauthenticated access is only possible from loopback.

2. **Loopback-only default**: When `CARRIER_SERVER_API_TOKEN` is unset, `cfg.Server.Host` is overridden to `127.0.0.1` regardless of configuration. Non-loopback without token still fatals.

3. **agentID validation**: Added `url.PathUnescape` for path-extracted IDs in `/api/status/` and `/api/logs/`. Added `validateAgentID` helper that rejects empty IDs, path separators (`/`, `\`), and characters outside `[a-zA-Z0-9._-]`. Applied to all agentID inputs (both path and body).

daemon/cmd/agentd/main.go
```diff
@@ -53,19 +63,300 @@ func main() {
...
+			}
+		}
+		lines, err := svc.Logs(agentID, tail)
+		if err != nil {
+			writeServiceError(w, err)
+			return
+		}
+		writeJSON(w, http.StatusOK, map[string]interface{}{"lines": lines})
+	})
+
```
Keith-CY: BS: `validateAgentID` currently allows `..` because the regex permits dots and there is no explicit `..` rejection. Please reject parent-directory tokens (e.g., `strings.Contains(id, "..")`) so path-derived IDs cannot carry traversal semantics.

dev01lay2:
Fixed the `..` rejection in `validateAgentID` (commit 168a1b6). Added `strings.Contains(id, "..")` check + `TestValidateAgentID` test covering `../`, `a..b`, empty, path separators, and invalid-start characters.

daemon/cmd/agentd/main.go
dev01lay2: Addressed — bearer token auth middleware is in place (CARRIER_SERVER_API_TOKEN). Without a token, binding is forced to loopback only. See `bearerAuthMiddleware` and `isLoopback` guard in current HEAD.

PR #446, opened 2026-02-14
fix: harden carrier 1h kanban board sync workflow

### LINKED ISSUE #445: [1h Kanban] Fix workflow runtime, hourly schedule, and issue-only snapshot
## Summary
Fix the Carrier 1h Kanban workflow to run reliably and match intended hourly behavior.

## Problems
- GraphQL client variable was used before being defined, causing runtime failure in discussion sync path.
- Schedule was every 30 minutes, which does not match one-hour cadence.
- Snapshot source included pull requests from the Issues API.
- Fallback issue number parsing lacked explicit validation.
- Timestamp label used a hardcoded timezone suffix.

## Scope
- Update .github/workflows/carrier-one-hour-kanban.yml only.

## Acceptance Criteria
- Workflow defines and uses a valid GitHub API client for GraphQL operations.
- Cron schedule is hourly (0 * * * *).
- Snapshot data excludes pull requests.
- Invalid CARRIER_1H_ISSUE_NUMBER fails fast with clear error.
- Timestamp label reflects configured timezone.

Keith-CY:
## Summary
- add robust GitHub client setup for GraphQL discussion operations
- switch cron cadence to hourly (0 * * * *) and align trigger label
- exclude pull requests from issue snapshot data
- validate fallback issue number configuration explicitly
- use configured timezone label in update header

## Verification
- ruby YAML parse check for .github/workflows/carrier-one-hour-kanban.yml

Closes #445

### DIFF
--- a/.github/workflows/carrier-one-hour-kanban.yml
+++ b/.github/workflows/carrier-one-hour-kanban.yml
@@ -1,0 +1,303 @@
+name: Carrier 1h Kanban Board Sync
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Report only, do not write content to discussion/issue'
+        required: false
+        default: 'false'
+  schedule:
+    - cron: '0 * * * *'
+
+permissions:
+  contents: read
+  issues: write
+  discussions: write
+
+jobs:
+  sync-1h-board:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update 1h Kanban snapshot
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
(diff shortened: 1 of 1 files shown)


dev01lay2 (approved):
CI is green, workflow logic is sound. Approving.

NBS: `classifyBucket` defaults to `Hotfix` for anything not matching explicit patterns — consider defaulting to a neutral bucket (e.g. `Uncategorized`) so new issues without labels don't silently appear in the hotfix lane.

NBS: The `discussions(first: 100)` query could miss the target discussion if the repo has >100. Consider filtering by title in the query or paginating.

NBS: REST calls (`github.paginate`, `github.rest.issues.*`) use the default `GITHUB_TOKEN` while GraphQL calls use `gh` (which may be `CARRIER_PROJECTS_TOKEN`). This is probably intentional for scope separation but worth a comment in the workflow for clarity.

PR #459, opened 2026-02-14
feat: bootstrap daemon HTTP API and phase-1 endpoint matrix

Keith-CY:
## Summary
- add daemon API contract doc with canonical path/method matrix and unified error-envelope mapping
- wire `agentd` HTTP bootstrap with route registry and graceful shutdown (`/healthz`, `/readyz`, `/api/v1/*`)
- implement phase-1 daemon API handlers for:
  - `GET /api/v1/agents`
  - `POST /api/v1/agents/{agent_id}/stop`
  - `POST /api/v1/agents/{agent_id}/upgrade`
  - `POST /api/v1/pairing/codes`
  - `POST /api/v1/pairing/verify-consume`
- add TTL pairing-code store with one-time consume semantics
- add unit tests for API handlers/store and route wiring
- ignore local worktree directories (`.worktree`, `.worktrees/`)

## Test Commands and Output Evidence
- `go test ./...` (from `daemon/`) failed in this environment with macOS dyld `missing LC_UUID` test-binary loader issue.
- `/bin/zsh -lc "CGO_ENABLED=0 go test ./..."` (from `daemon/`) passed:
  - `ok carrier/daemon/cmd/agentd`
  - `ok carrier/daemon/internal/api`
  - `ok carrier/daemon/internal/lifecycle`
  - and all remaining daemon packages `ok`

Closes #334
Closes #385
Closes #386
Closes #389
Closes #394
Closes #405
Closes #407

PR #529, opened 2026-02-14
feat(daemon): rebase PR 459 on main and merge API v1 implementation

Keith-CY:
This PR is a rebase-on-main follow-up for PR #459 with merge conflict resolution.

Changes:
- Re-apply API /phase1 contract and lifecycle endpoint work from PR 459 on top of latest main.
- Resolve conflicts by preserving /api/v1 behavior with back-compat for older clients.
- Update daemon agentd startup and tests to reflect merged API contract behavior.
- Retain required security and migration guards for new lifecycle endpoints.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ docs/Agent_Installation_Platform_PRD_extra.md
 node_modules/
 daemon/coverage.html
 daemon/coverage.out
+.worktree
+.worktrees/
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -49,6 +49,10 @@ The gateway is a Bun/TypeScript WebSocket server that:
 
 Typed command definitions shared between gateway and clients, defining the request/response schema for each operation (`/pair`, `/agents`, `/install`, `/start`, `/stop`, `/status`, `/logs`, `/upgrade`, `/diagnose`).
 
+### Daemon API Contract (`docs/daemon-api-contract.md`)
+
+Canonical endpoint/method matrix and error-envelope mapping for daemon HTTP APIs consumed by the gateway.
+
 ## Data Flow
 
 1. Client connects to gateway via WebSocket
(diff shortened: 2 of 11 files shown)


dev01lay2 (changes requested):
Solid refactor — extracting the API layer into its own package with proper error envelopes, health server, and pairing store is a big improvement. One blocking issue on input validation.

**BS:** `parseAgentActionPath` extracts `agentID` from the URL path without URL-decoding or input validation. This is the same issue flagged in #459 — encoded path traversal characters (`%2F`, `%2E%2E`) can bypass the simple string split. Please add `url.PathUnescape` + validation (reject `/`, `\`, `..`, empty, enforce `^[a-zA-Z0-9][a-zA-Z0-9._-]*$`) before passing to lifecycle methods.

NBS: `readJSON` uses `DisallowUnknownFields` — this could break forward compatibility if clients send fields the server does not yet know. Consider removing it for more resilient versioning.

NBS: Only `stop` and `upgrade` are wired in `handleAgentAction`; the contract doc lists install/start/status/logs/diagnose as planned. Might be worth returning 501 (Not Implemented) instead of 404 for documented-but-unimplemented actions to distinguish "wrong URL" from "not yet built."

Keith-CY:
BS: `parseAgentActionPath` still accepts raw `%2F`/%2E%2E` style path segments. Please URL-decode (`url.PathUnescape`) and validate agentId before passing to lifecycle methods to prevent malformed traversal and ambiguous matching.

Keith-CY:
BS: Still unresolved compatibility blockers in daemon API: /daemon/internal/api/server.go only handles action endpoints stop/upgrade, but in-repo clients still call install/start/status/logs/diagnose under /api/v1/agents/{agentId}/... causing runtime 404s. Add compatibility handlers (or explicit migration plan + client updates) before merge.

Keith-CY:
NBS: `daemon/internal/api/pairing_store.go` has TTL and `CleanupExpired()`, but cleanup is never invoked from request flow in `daemon/internal/api/server.go`. Under repeated `POST /api/v1/pairing/codes` usage, expired entries can accumulate unbounded in memory. Consider opportunistic cleanup on issue/verify or a periodic sweeper.

PR #534, opened 2026-02-14
fix(gateway): add path validation to defaultReadFile

### LINKED ISSUE #474: [review-followup] PR #471: `defaultReadFile` trusts `fileRef` from the token store without path-componen...
Source PR: https://github.com/Keith-CY/carrier/pull/471
PR title: feat: add gateway server runtime and HTTP daemon client path

## Suggestion
- [ ] `defaultReadFile` trusts `fileRef` from the token store without path-component validation; if a future caller passes user-influenced paths, this could become a traversal vector. Current usage is safe since `issue()` is server-controlled.

## Trace
- review:dev01lay2

dev01lay2:
Fixes #474

Adds path validation to defaultReadFile to prevent potential path traversal attacks. The function now rejects relative paths and paths containing .. components.

**Changes:**
- Added path validation to reject non-absolute paths and .. components
- Added security warning log for rejected paths
- Added test coverage for path traversal rejection

**Security:**
Current usage is safe (server-controlled paths via token store), but this guards against future misuse if callers pass user-influenced paths.

**Testing:**
- ✅ All existing tests pass
- ✅ New test verifies path traversal rejection

### DIFF
--- a/gateway/src/server.test.ts
+++ b/gateway/src/server.test.ts
@@ -161,6 +161,6 @@ describe("gateway runtime routes", () => {
     
     // Should get 404 because readFile rejects the path
     expect(response.status).toBe(404);
-    expect(capturedPath).toBe(maliciousPath);
+    expect(capturedPath as string | null).toBe(maliciousPath);
   });
 });


Keith-CY:
NBS: `gateway/src/server.ts` path guard in `defaultReadFile` currently requires `fileRef.startsWith("/")`. That is Unix-specific and rejects valid Windows absolute paths (for example `C:\\...`). If cross-platform runtime behavior is required, please switch to platform-aware absolute-path validation/normalization.

PR #541, opened 2026-02-14
test(gateway): cover actor/request-id header propagation in HttpDaemonClient

### LINKED ISSUE #476: test(gateway): cover actor/request-id header propagation in HttpDaemonClient
## Scope
Add focused unit tests for `gateway/src/daemon/http_client.ts` request metadata forwarding, to verify actor/request-id headers are consistently attached across lifecycle calls.

## Why
The new HTTP daemon client is now runtime-critical. We already test response parsing and error mapping, but not request metadata propagation, which is needed for traceability and audit consistency.

## Acceptance Criteria
- Add tests in `gateway/src/daemon/http_client.test.ts` (or a nearby dedicated test file) covering at least:
  - `x-carrier-actor` and `x-carrier-request-id` propagation for `listAgents`.
  - Same header propagation for one mutating call (`startAgent` or `stopAgent`).
- Tests should assert exact header keys/values sent to `fetch`.
- `cd gateway && bun test` passes.

dev01lay2:
Closes #476

### DIFF
--- a/gateway/src/daemon/http_client.test.ts
+++ b/gateway/src/daemon/http_client.test.ts
@@ -94,10 +94,10 @@ describe("HttpDaemonClient", () => {
 describe("HttpDaemonClient header propagation", () => {
   test("listAgents propagates actor and request-id headers", async () => {
     let capturedHeaders: HeadersInit | undefined;
-    const fetchMock: typeof fetch = async (input, init) => {
+    const fetchMock = (async (input: RequestInfo | URL, init?: RequestInit) => {
       capturedHeaders = init?.headers;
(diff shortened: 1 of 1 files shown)


Akane-CN (commented):
Re-review on current HEAD completed.

The added header propagation tests are directionally good and cover key command paths (`list/start/stop/install`).

I’m not approving yet because required checks are not green on this head (`Gateway Check` failed and `sync` failed; daemon tests still pending at check time).

NBS: Consider asserting via a normalized `Headers` object (e.g., `new Headers(capturedHeaders)`) instead of casting to `Record<string, string>`, to avoid shape differences across fetch implementations.

PR #715, opened 2026-02-15
test(agentd): add bearerAuthMiddleware auth/scope coverage

### LINKED ISSUE #714: test(agentd): add focused coverage for bearerAuthMiddleware auth/scope contracts
## Problem
`daemon/cmd/agentd/main_test.go` currently verifies health/API mounting and shutdown helpers, but it does not cover `bearerAuthMiddleware` behavior at all. That leaves important auth-path contracts untested (API path gating vs health exemptions, missing header behavior, and valid bearer pass-through).

## Why it matters
`bearerAuthMiddleware` is the last guard before daemon API routes. Without focused tests, future refactors can silently weaken auth enforcement or accidentally break non-auth endpoints (`/healthz`, `/readyz`). These regressions would be security-impacting or operationally noisy.

Add table-driven unit tests in `daemon/cmd/agentd/main_test.go` (or a dedicated `_test.go` in the same package) that exercise middleware behavior with a stub next handler:
- `/api/...` with no `Authorization` header returns `401` JSON error.
- `/api/...` with wrong bearer token returns `401`.
- `/api/...` with exact expected bearer token calls next handler and returns downstream status.
- non-API paths (e.g., `/healthz`, `/readyz`) bypass auth checks and call next handler even without token.
- assert response content-type for unauthorized path remains JSON.

## Acceptance criteria
- [ ] New tests cover both allow and deny paths of `bearerAuthMiddleware`.
- [ ] Tests assert API-only enforcement scope (auth required only for `/api/` prefix).
- [ ] Tests assert unauthorized responses are `401` with JSON content-type.
- [ ] Tests are deterministic and avoid external network/process dependencies.

- `cd daemon && go test ./cmd/agentd -run TestBearerAuthMiddleware -count=1`

Akane-CN:
## Summary
- add table-driven tests for `bearerAuthMiddleware` in `daemon/cmd/agentd/main_test.go`
- cover unauthorized API access (missing/wrong bearer token)
- cover authorized API pass-through and non-API bypass behavior (`/healthz`, `/readyz`)
- assert unauthorized responses return JSON content-type and expected body

## Validation
- `cd daemon && go test ./cmd/agentd -count=1`

Closes #714

### DIFF
--- a/daemon/cmd/agentd/main_test.go
+++ b/daemon/cmd/agentd/main_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dev01lay2 (commented):
Clean table-driven tests, good coverage of the auth/bypass paths. One minor suggestion below.

PR #752, opened 2026-02-15
fix(daemon): stop lifecycle agents by process group

### LINKED ISSUE #712: bug(daemon): Stop() signals only parent PID, not process group; child processes can outlive agent
## Problem
`daemon/internal/lifecycle/process.go` starts agents with `SysProcAttr.Setpgid = true`, but `Stop()` sends signals only to `info.cmd.Process` (single PID), not to the process group.

If an agent spawns child processes, stopping the parent can leave descendants running. This creates orphaned background work and can keep ports/files busy after the daemon reports a successful stop.

## Impact / Risk
- Reliability: `Stop` can report success while child processes continue running.
- Operational risk: lingering processes can interfere with restarts/upgrades and consume resources unexpectedly.
- Correctness: process-group setup is currently underused, so shutdown semantics are inconsistent with intent.

## Proposed change
Use process-group signaling when stopping:
- Send `SIGTERM` to `-pid` (process group) when `Setpgid` is enabled.
- After timeout, send `SIGKILL` to the same process group.
- Keep current fallback behavior for platforms/configs where pgid signaling is unavailable.
- Add explicit logging/metrics that distinguish parent-exit vs full-group-exit behavior.

## Acceptance criteria
- [ ] `Stop()` targets the spawned process group (not only parent PID) when available.
- [ ] Grace period behavior remains: TERM first, KILL after timeout.
- [ ] No child process remains after successful stop in regression test scenarios.
- [ ] Existing single-process stop behavior remains green.

- Add a lifecycle test binary/script that spawns a child process that outlives parent unless group-signaled.
- Verify `Stop()` terminates both parent and child for graceful and forced-stop paths.
- Run `go test ./...` in `daemon/` and ensure no regressions in existing lifecycle tests.

Akane-CN:
## Summary
- stop lifecycle-managed agents by signaling the **process group** (`-pid`) instead of only the parent process
- apply the same process-group targeting for SIGKILL fallback after timeout
- add regression test proving child processes are terminated when stopping an agent

## Validation
- `cd daemon && go test ./internal/lifecycle -run TestProcessManager_StopKillsChildProcessesViaProcessGroup -count=1 -v`
- `cd daemon && go test ./internal/lifecycle -run TestProcessManager_StartStop -count=1`

Closes #712

### DIFF
--- a/daemon/internal/api/server_test.go
+++ b/daemon/internal/api/server_test.go
@@ -65,7 +65,7 @@ func sampleManifest() manifest.Manifest {
 			Type:    manifest.RuntimeTypeLocalBinary,
 			Install: manifest.CommandSpec{Command: "install-openclaw"},
 			Upgrade: manifest.CommandSpec{Command: "upgrade-openclaw"},
-			Start:   manifest.CommandSpec{Command: "start-openclaw"},
+			Start:   manifest.CommandSpec{Command: "tail -f /dev/null"},
 			Stop:    manifest.CommandSpec{Command: "stop-openclaw"},
 		},
 		Network: manifest.NetworkSpec{


dev01lay2 (commented):
Clean and correct. `Setpgid: true` is already set in `Start()`, so `syscall.Kill(-pid, ...)` targets the right process group. ESRCH guard handles the race where the process exits between the check and the signal. Test is solid — spawns a real child, confirms it dies with the group.

No issues found.

Akane-CN:
Pushed follow-up lifecycle fix series for failing checks:
- aligned process-manager err handling used by lint checks.
- added immediate start-failure detection for process mode.
- kept process-group stop semantics to avoid child-process hang on stop.
- validated key lifecycle crash-loop tests locally after updates.

Akane-CN:
Additional CI follow-up pushed:
- API lifecycle tests now use a long-running start command (`tail -f /dev/null`) so endpoint contract tests align with process-mode runtime behavior.
- lifecycle crash-loop post-upgrade path was also validated locally after recent updates.

Issue #625, opened 2026-02-14
bug(daemon): no PAIR_CODE printed on startup — README flow broken


dev01lay2:
## Problem

README step 6 says:
> Get your pairing code from the daemon terminal output (it is generated when daemon/gateway pairing flow is ready).

But `agentd` does **not** print any pairing code on startup. Actual output:

```
agentd scaffold booted (listen=127.0.0.1:9090 log=info/text)
catalog:
- NanoClaw (nanoclaw): candidate
- OpenClaw (openclaw): active
agentd HTTP server listening on 127.0.0.1:9090
```

The daemon has a `/api/v1/pairing/codes` endpoint that can generate codes, but nothing calls it at startup.

## Expected
On startup, `agentd` should auto-generate a pairing code and print it to terminal, e.g.:
```
PAIR_CODE: pair-abc123 (expires in 10 minutes)
```

## Impact
Non-technical users following the README cannot complete the pairing step without developer knowledge of the API.

## Self-check context
Discovered during automated README flow validation on 2026-02-14.

Akane-CN:
Quick sweep note: PR #764 addresses this startup PAIR_CODE gap, but it currently includes a committed binary (`daemon/agentd`) which is blocking review.

If that artifact is removed and CI stays green, the functional fix path here looks ready to land.

PR #768, opened 2026-02-15
feat(gateway): Verify pair codes against daemon API

### LINKED ISSUE #599: Gateway /pair does not verify codes against daemon API
## Problem

During README self-check, `/pair <code>` via gateway always returns `E_PAIR_CODE_INVALID`, even with a valid daemon-issued code.

## Root cause

Gateway's `/pair` handler checks `SessionStore.pairCodes` (an in-memory map), but daemon-issued codes (from `POST /api/v1/pairing/codes`) are never registered there. The gateway has no bridge to call `POST /api/v1/pairing/verify-consume` on the daemon.

Tests work because they manually call `deps.sessions.registerPairCode()`.

## Expected behavior

When a user sends `/pair <code>`, the gateway should call `POST /api/v1/pairing/verify-consume` on the daemon to validate and consume the code, then create the local session on success.

## Impact

**Blocks the entire install flow** — without pairing, no subsequent command (`/agents`, `/install`, etc.) can execute via gateway.

## Reproduction

```bash
# Start daemon + gateway
./agentd &
cd gateway && bun run dev &

# Get code from daemon
CODE=$(curl -s -X POST http://127.0.0.1:9090/api/v1/pairing/codes | jq -r .code)

# Try pairing via gateway — fails
curl -s -X POST http://127.0.0.1:8787/command \
  -H 'Content-Type: application/json' \
  -d "{\"input\":\"telegram chat1 r1 /pair $CODE\"}"
# Returns: E_PAIR_CODE_INVALID
```

dev01lay2:
## Summary
Move pairing code verification from gateway's local SessionStore to daemon API, establishing daemon as the source of truth for pairing.

### DaemonClient Interface
- Add `verifyPairCode(code: string, ctx: RequestContext): Promise<void>` method
- Implement in HttpDaemonClient: POST /api/v1/pairing/verify-consume
- Implement in InMemoryDaemonClient with test helper `registerPairCode()`

### Gateway /pair Handler
- Call `daemon.verifyPairCode()` before creating session
- Handle DaemonClientError responses (invalid/expired codes)
- Create session using new `SessionStore.createSession()` method

### SessionStore
- Add `createSession()` method for direct session creation
- Deprecate local pair code validation (now handled by daemon)

### Testing
- ✅ All 200 tests pass
- Updated all tests to register codes with daemon instead of session store
- E2E tests verify full daemon → gateway pairing flow

## Breaking Changes
Gateway now requires daemon /api/v1/pairing/verify-consume endpoint (added in #625)

Fixes #599

### DIFF
--- a/gateway/src/command-routing.test.ts
+++ b/gateway/src/command-routing.test.ts
@@ -15,8 +15,10 @@ function buildDeps(): GatewayDependencies {
 }
 
 function pairChat(d: GatewayDependencies, provider: "telegram" | "discord" | "feishu" = "telegram", chatId = "100"): void {
-  d.sessions.registerPairCode("pair-ok", 300);
-  d.sessions.pair({ provider, chatId, code: "pair-ok" });
+  if (d.daemon instanceof InMemoryDaemonClient) {
+    d.daemon.registerPairCode("pair-ok");
+  }
+  d.sessions.createSession({ provider, chatId });
 }
 
 describe("command routing: /pair", () => {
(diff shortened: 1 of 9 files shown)


Akane-CN (approved):
LGTM. I reviewed the current HEAD commit and CI is green.

No blocking issues found.

PR #772, opened 2026-02-15
fix(daemon): call Cleanup on shutdown to stop managed processes

### LINKED ISSUE #746: bug(daemon): shutdown path never calls ProcessManager.Cleanup — orphan processes on failed stop
## Problem

In `daemon/cmd/agentd/main.go`, when the daemon shuts down, `stopAllAgents()` iterates through agents and calls `svc.Stop()` for each. However, if `svc.Stop()` fails, the underlying `ProcessManager.Cleanup()` is never invoked as a fallback.

```go
func stopAllAgents(svc *lifecycle.Service) error {
    agents := svc.ListAgents()
    var firstErr error
    for _, agent := range agents {
        if agent.Runtime == lifecycle.RuntimeStateRunning {
            if err := svc.Stop(context.Background(), agent.ID); err != nil {
                // Error logged, but process may still be running
                // ProcessManager.Cleanup() is never called
            }
        }
    }
    return firstErr
}
```

The `ProcessManager` has a `Cleanup()` method specifically designed for graceful shutdown, which sends SIGTERM→SIGKILL to all tracked processes. But it's never called from any shutdown path.

## Impact
- If `svc.Stop()` fails (e.g., state inconsistency, lock contention), orphaned child processes survive daemon shutdown
- On daemon restart, these orphaned processes may cause port conflicts or resource contention
- The `Cleanup()` method exists but is dead code in the shutdown path

## Suggested Fix
Call `ProcessManager.Cleanup()` at the end of shutdown, or expose it through the `Service` type:

```go
// In main.go shutdown sequence:
if err := shutdownAgents(svc, shutdownTimeout); err != nil {
    // Even if graceful stop failed, force-cleanup all processes
    svc.ProcessManager().Cleanup()
    log.Fatalf("shutdown error: %v", err)
}
```

## Files
- `daemon/cmd/agentd/main.go` — `stopAllAgents()`
- `daemon/internal/lifecycle/process.go` — `Cleanup()` (never called from main)
- `daemon/internal/lifecycle/service.go` — `processManager` field (unexported, inaccessible from main)

dev01lay2:
Closes #746

### DIFF
--- a/daemon/cmd/agentd/main.go
+++ b/daemon/cmd/agentd/main.go
@@ -159,6 +159,8 @@ func stopAllAgents(svc *lifecycle.Service) error {
 			}
 		}
 	}
+	// Cleanup any remaining managed processes
+	svc.Cleanup()
 	return firstErr
 }
 
--- a/daemon/internal/lifecycle/service.go
+++ b/daemon/internal/lifecycle/service.go
@@ -434,6 +434,13 @@ func (s *Service) DiagnosisHandoffs() []DiagnosisHandoff {
 	return out
 }
 
+// Cleanup stops all managed processes for graceful shutdown.
+func (s *Service) Cleanup() {
+	if s.processManager != nil {
+		s.processManager.Cleanup()
+	}
+}
+
 func (s *Service) CleanupExpiredDiagnosisHandoffs() int {
 	cutoff := s.now().Add(-s.handoffTTL)
 


Akane-CN (approved):
LGTM. I reviewed the current HEAD commit and CI is green.

No blocking issues found.

PR #775, opened 2026-02-15
fix(gateway): add cleanup for expired rate limit windows

### LINKED ISSUE #755: bug(gateway): RateLimiter sessionWindows map never evicts disconnected sessions — unbounded memory growth
## Problem

In `gateway/src/ratelimit/index.ts`, the `RateLimiter` class stores per-session timestamp arrays in a `Map<string, number[]>`. While timestamps within each session are pruned on every `check()` call, the map entries themselves are never removed.

`clearSession()` exists but is never called by any code path. Over time, every unique `provider:chatId` pair that ever sends a command creates a permanent map entry. For a long-running gateway, this is an unbounded memory leak.

## Impact

On a gateway running for weeks/months with many unique chat sessions, memory usage grows monotonically. The global `globalWindow` array is pruned correctly, but each session key persists with an empty `number[]` array indefinitely.

## Suggested Fix

1. Add periodic cleanup in the `check()` method: after pruning a session's timestamps, if the array is empty, delete the map entry.
2. Alternatively, add a `startPeriodicCleanup()` method (similar to `DownloadTokenStore`) that sweeps empty/stale sessions every N minutes.

## Files
- `gateway/src/ratelimit/index.ts` (sessionWindows map, check method)

dev01lay2:
Closes #755

### DIFF
--- a/gateway/src/ratelimit/index.ts
+++ b/gateway/src/ratelimit/index.ts
@@ -27,6 +27,7 @@ export class RateLimiter {
   private readonly config: RateLimitConfig;
   private readonly sessionWindows: Map<string, number[]> = new Map();
   private globalWindow: number[] = [];
+  private cleanupIntervalId: Timer | null = null;
 
   constructor(config?: Partial<RateLimitConfig>) {
     this.config = {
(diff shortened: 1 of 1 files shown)


Akane-CN (commented):
Thanks for the update. I reviewed the current HEAD, but required checks are still pending, so I’m holding approval for now.

Once CI is fully green, I can re-check quickly.

dev01lay2:
CI is fully green now — all checks pass. Ready for re-review when you have a moment. 🐰

dev01lay2:
CI is all green now ✅ — ready for review whenever you have a moment.

PR #780, opened 2026-02-15
fix(health): limit /healthz to status only

### LINKED ISSUE #754: security(daemon): /healthz endpoint exposes version, uptime, and agent count without authentication
## Problem

The `/healthz` endpoint in `daemon/internal/health/health.go` returns sensitive operational metadata without requiring any authentication:

```json
{
  "status": "ok",
  "uptime": "4h32m10s",
  "uptime_seconds": 16330,
  "running_agents": 3,
  "version": "0.2.1"
}
```

In `daemon/cmd/agentd/main.go`, the `bearerAuthMiddleware` only protects `/api/` routes, explicitly leaving `/healthz` and `/readyz` exempt (lines 114-121).

While health endpoints are commonly unauthenticated, this one returns more than a simple OK/NOT-OK status. An attacker with network access can fingerprint the exact daemon version, determine how long it has been running (useful for timing attacks around restarts/upgrades), and enumerate how many agents are active.

## Suggested Fix

Option A: Reduce the unauthenticated response to just `{"status":"ok"}` and require auth for the detailed fields.

Option B: Add a `?verbose=true` query parameter that requires the bearer token to return extended info.

## Files
- `daemon/internal/health/health.go` (handleHealthz)
- `daemon/cmd/agentd/main.go` (bearerAuthMiddleware exemption)

dev01lay2:
Closes #754

### DIFF
--- a/daemon/internal/health/health.go
+++ b/daemon/internal/health/health.go
@@ -44,14 +44,6 @@ func (s *Server) Handler() *http.ServeMux {
 	return mux
 }
 
-type healthzResponse struct {
-	Status        string `json:"status"`
-	Uptime        string `json:"uptime"`
-	UptimeSeconds int64  `json:"uptime_seconds"`
-	RunningAgents int    `json:"running_agents"`
-	Version       string `json:"version"`
-}
-
 func (s *Server) handleHealthz(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


Akane-CN (approved):
LGTM. I reviewed the latest commit and verified checks are green.

PR #783, opened 2026-02-15
fix(gateway): add symlink traversal protection

### LINKED ISSUE #706: security(gateway): defaultReadFile path traversal via symlink — no realpath resolution
## Problem

`defaultReadFile` in `gateway/src/server.ts` guards against `/../` and trailing `/..` in the `fileRef` string, but does **not** resolve symlinks. If an attacker can create a symlink at a valid-looking path (e.g. via a compromised agent's diagnose artifact or log directory), the symlink could point to an arbitrary file outside the intended directory.

```ts
// Current check (server.ts)
if (!fileRef.startsWith("/") || fileRef.includes("/../") || fileRef.endsWith("/..")) {
  console.warn(`[security] rejected potentially unsafe fileRef: ${fileRef}`);
  return null;
}
```

## Impact

An attacker who can influence the `fileRef` stored in a download token (e.g. through a crafted diagnose artifact path) could read arbitrary files from the host via the `/downloads/` endpoint.

## Suggested Fix

After the string-level check, resolve the path with `Bun.file(fileRef).name` or Node's `fs.realpathSync` equivalent and verify the resolved path is still within an allowed directory (e.g. `/tmp` or the daemon's data directory).

```ts
import { realpathSync } from 'node:fs';

const resolved = realpathSync(fileRef);
const ALLOWED_PREFIXES = ['/tmp/', '/var/lib/carrier/'];
if (!ALLOWED_PREFIXES.some(p => resolved.startsWith(p))) {
  console.warn(`[security] resolved path outside allowed dirs: ${resolved}`);
  return null;
}
```

## File
- `gateway/src/server.ts` — `defaultReadFile` function

dev01lay2:
Closes #706

### DIFF
--- a/gateway/src/server.test.ts
+++ b/gateway/src/server.test.ts
@@ -3,6 +3,9 @@ import { InMemoryDaemonClient } from "./daemon/client";
 import { HttpDaemonClient } from "./daemon/http_client";
 import { DownloadTokenStore } from "./downloads/token_store";
 import { SessionStore } from "./session/store";
+
+// Tests create temporary files under /tmp; allow defaultReadFile to serve them.
+process.env.ARTIFACT_ROOT = "/tmp";
 import {
   composeMiddleware,
   createGatewayRuntime,
--- a/gateway/src/server.ts
+++ b/gateway/src/server.ts
@@ -225,29 +225,39 @@ function buildContentDisposition(filename: string): string {
 }
 
 async function defaultReadFile(fileRef: string): Promise<Blob | null> {
+  const { resolve: pathResolve, sep } = await import("node:path");
+  const { realpath: fsRealpath } = await import("node:fs/promises");
+
   // Validate path to prevent traversal attacks - reject relative paths and .. components
   // Current callers are server-controlled, but this guards against future misuse
   if (!fileRef.startsWith("/") || fileRef.includes("/../") || fileRef.endsWith("/..")) {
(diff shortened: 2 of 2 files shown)


Akane-CN (changes requested):
Thanks for the security hardening pass. I found one blocking issue in the current symlink check.

dev01lay2:
Pushed fix in 8afceca — now using `fs.realpath` to resolve symlinks and validating the resolved path stays within the parent directory before serving the file. This should address the traversal concern. Ready for re-review. 🐰

dev01lay2:
CI is fully green now ✅ (all checks pass on latest `7f006bb`). The additional commit since your last look is a race-condition fix in `Stop()` — marks state as Stopped before killing the process to prevent `monitorProcess` from overwriting it to Crashing. Ready for re-review when you get a chance!

dev01lay2:
CI is green on the latest commit (`d6f0f39`). The `ARTIFACT_ROOT` boundary confinement is now implemented — both the canonical path and the symlink-resolved path are checked against `rootPrefix` via `startsWith`. This should address the blocking concern about arbitrary absolute-path reads (e.g. `/etc/hosts`).

Ready for re-review when you have a moment @Akane-CN.

Akane-CN (approved):
Re-review on current HEAD `d6f0f39`.

CI is green, and the artifact-root boundary + canonical path checks address the previously blocking path-escape gap.

Approving.

PR #784, opened 2026-02-15
fix(gateway): add graceful shutdown handlers

### LINKED ISSUE #707: bug(gateway): no graceful shutdown — active downloads and in-flight commands lost on SIGTERM
## Problem

The gateway entry point in `server.ts` calls `Bun.serve()` but has **no signal handling** for SIGTERM/SIGINT. When the process is killed (e.g. by a container orchestrator, systemd, or Ctrl-C), in-flight `/command` requests and active `/downloads/` transfers are abruptly terminated with no draining period.

```ts
// server.ts — current code
if (import.meta.main) {
  const server = startGatewayServer();
  console.log(`gateway server listening on http://${server.hostname}:${server.port}`);
  // ← no signal handling, no graceful shutdown
}
```

In contrast, the daemon (`main.go`) correctly handles SIGTERM/SIGINT with a configurable shutdown timeout and drains HTTP connections.

## Impact

- Users may see partial/corrupt responses when the gateway restarts
- Download tokens consumed but content not fully delivered are wasted (token is single-use)
- No log message indicating shutdown, making debugging harder

## Suggested Fix

```ts
if (import.meta.main) {
  const server = startGatewayServer();
  console.log(`gateway server listening on http://${server.hostname}:${server.port}`);

  const shutdown = () => {
    console.log('shutdown signal received, draining connections...');
    server.stop(true); // graceful: waits for in-flight requests
    process.exit(0);
  };
  process.on('SIGTERM', shutdown);
  process.on('SIGINT', shutdown);
}
```

## File
- `gateway/src/server.ts` — `import.meta.main` block

dev01lay2:
Closes #707

### DIFF
--- a/daemon/internal/health/health.go
+++ b/daemon/internal/health/health.go
@@ -44,14 +44,6 @@ func (s *Server) Handler() *http.ServeMux {
 	return mux
 }
 
-type healthzResponse struct {
-	Status        string `json:"status"`
-	Uptime        string `json:"uptime"`
-	UptimeSeconds int64  `json:"uptime_seconds"`
-	RunningAgents int    `json:"running_agents"`
-	Version       string `json:"version"`
-}
-
 func (s *Server) handleHealthz(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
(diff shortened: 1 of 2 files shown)


Akane-CN (approved):
LGTM. I reviewed the latest commit and verified checks are green.

PR #859, opened 2026-02-15
fix(daemon/api): reject trailing JSON payloads in readJSON

### LINKED ISSUE #789: bug(daemon/api): readJSON accepts trailing JSON payload after first object
## Problem
`daemon/internal/api/readJSON` decodes exactly one JSON value but does not assert end-of-input after the first decode. As a result, payloads like `{ "code": "pair-a" } { "code": "pair-b" }` are accepted instead of rejected.

## Why it matters
- API behavior becomes ambiguous for malformed/multi-value request bodies.
- Different clients/proxies may interpret trailing bytes differently, increasing parser differential risk.
- This weakens request validation guarantees that other API hardening work relies on.

1. Refactor `readJSON` to enforce a **single JSON value** contract:
   - decode once into destination
   - decode again into `struct{}` (or token) and require `io.EOF`
   - return `E_USAGE`-mapped error when trailing non-whitespace content exists
2. Keep current size-limit and unknown-field behavior.
3. Add focused API tests that currently fail, then pass after fix:
   - `/api/v1/pairing/verify-consume` rejects multi-value JSON body
   - `/api/v1/pairing/codes` rejects valid first object with trailing junk

## Acceptance criteria
- [ ] Requests with trailing JSON values or non-whitespace garbage are rejected with `400` and `E_USAGE`.
- [ ] Requests with a single valid JSON value continue to work unchanged.
- [ ] Existing strict unknown-field behavior remains intact.
- [ ] New regression tests cover at least two endpoints using `readJSON`.

- `cd daemon && go test ./internal/api -run 'Test(ReadJSON|VerifyConsumePairCode|IssuePairCode)'`

Akane-CN:
## Summary
- enforce `readJSON` single-value contract by requiring EOF after first decode
- keep existing request-size limit and `DisallowUnknownFields` strictness
- add regression tests to reject trailing payloads on both `/api/v1/pairing/verify-consume` and `/api/v1/pairing/codes`

## Validation
- `cd daemon && go test ./internal/api -run 'Test(ReadJSON|VerifyConsumePairCode|IssuePairCode)'`

Closes #789

### DIFF
--- a/daemon/internal/api/server.go
+++ b/daemon/internal/api/server.go
@@ -423,6 +423,12 @@ func readJSON(r *http.Request, dst any) error {
 	if err := dec.Decode(dst); err != nil {
 		return fmt.Errorf("invalid json: %w", err)
 	}
+
+	var trailing struct{}
+	if err := dec.Decode(&trailing); err != io.EOF {
+		return fmt.Errorf("invalid json: trailing content after first json value")
+	}
+
 	return nil
 }
 
--- a/daemon/internal/api/server_test.go
+++ b/daemon/internal/api/server_test.go
@@ -94,6 +94,11 @@ func doJSONRequest(t *testing.T, handler http.Handler, method, path string, body
 		payload = raw
 	}
 
+	return doRawJSONRequest(t, handler, method, path, payload)
+}
+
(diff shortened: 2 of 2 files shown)


dev01lay2 (approved):
Clean fix — trailing-value rejection in `readJSON` is correct and well-tested. LGTM.

PR #871, opened 2026-02-15
feat(catalog): enforce checksum verification in install script

### LINKED ISSUE #851: [Micro][T059][#17] Enforce checksum verification in install/upgrade scripts.
## Micro Task
Enforce checksum verification in install/upgrade scripts.

## Metadata
- Parent Decomposition: #17
- Roll-up Tracker: #119
- Task Code: `T059`
- Owner: @dev01lay2
- Size: `S`
- WIP Policy: max 2 active tasks per owner

## Dependencies
- Human: `T058`
- depends-on: #850

## DoR (Definition of Ready)
- [ ] Exact file/function scope is listed in first implementation comment.
- [ ] Expected API/behavior contract is stated (or linked).
- [ ] Target tests/specs are identified before code change.
- [ ] All `depends-on` issues are closed.
- [ ] For API/behavior tasks: spec-first required = `optional`.

## DoD (Definition of Done)
- [ ] Exactly one functional change aligned to this micro task.
- [ ] Focused tests/spec updates are included and passing.
- [ ] Docs are updated when API/contract behavior changes.
- [ ] CI is green on the PR.
- [ ] PR body includes evidence links and references parent/tracker issues.

## Parent Closure Contract
- Parent issue closes only after all required micro tasks are closed and evidence links are present.

dev01lay2:
Closes #851

### DIFF
--- a/daemon/internal/catalog/openclaw-installer_test.sh
+++ b/daemon/internal/catalog/openclaw-installer_test.sh
@@ -1,52 +1,55 @@
 #!/bin/bash
 # Test script for openclaw-installer.sh checksum verification
+# All tests run the installer in a subshell so that `exit` inside the
+# script does not kill the test runner.
 
 set -e
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 INSTALLER="$SCRIPT_DIR/openclaw-installer.sh"
+PASS=0
+FAIL=0
+
+pass() { echo "PASS: $1"; PASS=$((PASS + 1)); }
+fail() { echo "FAIL: $1"; FAIL=$((FAIL + 1)); }
 
 echo "Testing openclaw-installer.sh checksum verification..."
 
-# Test 1: Verify script fails when no checksum is provided
+# Test 1: No checksum at all → must exit non-zero
 echo "Test 1: Should fail when no checksum provided"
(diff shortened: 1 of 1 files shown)


Keith-CY (commented):
NBS: This review is non-blocking. Findings below with `NBS:` prefix.

dev01lay2:
Self-review notes:

- CI all green (shellcheck + daemon lint/tests + gateway check ✅)
- The `OPENCLAW_CHECKSUM` env-var fallback allows standalone use without changing the daemon call path ($2 still works)
- **Minor**: Test 3/4 use `. $INSTALLER` (source) which means an `exit 1` inside the installer kills the test runner — subshell (`bash $INSTALLER`) would be safer, but not critical since these tests aren't in CI

Ready for review @Keith-CY

Keith-CY (approved):
Looks good overall. The checksum fallback rewrite and the shell tests are now robust and there are no blocking issues in this revision.

daemon/internal/catalog/openclaw-installer_test.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
Keith-CY: NBS: Could you hook this into CI/test entrypoints (e.g., `test:` target in Makefile) so this script runs automatically? Right now it only runs when manually invoked.

PR #884, opened 2026-02-15
security(gateway): validate session token on all command requests

### LINKED ISSUE #728: security(gateway): session token issued at pairing is never validated on subsequent commands
## Problem

After a successful `/pair` command, the gateway issues a `sessionToken` in the response but **never checks it on subsequent requests**. The `handleCommand` function only verifies that a session exists for `provider:chatId` — it never requires or validates the session token.

**File:** `gateway/src/index.ts` — lines around `handleCommand`, specifically:
```ts
const session = deps.sessions.getSession(cmd.provider, cmd.chatId);
if (!session) {
  return errorResponse(...);
}
```

This lookup is purely by provider+chatId. Any client that knows (or guesses) a provider+chatId pair can issue commands without the session token.

## Impact
- The session token provides no security value — it's issued but ignored.
- If the gateway `/command` endpoint is exposed (see #631), any caller can impersonate a paired chat.
- Violates the principle of least authority — the pairing flow suggests token-based auth but doesn't enforce it.

## Proposed Fix
Require `sessionToken` as a field in the command input (or as a Bearer header), and validate it matches the stored session before dispatching any command.

## Acceptance Criteria
- [ ] Commands after `/pair` must include the session token.
- [ ] Mismatched or missing token returns `E_SESSION_TOKEN_INVALID`.
- [ ] Existing tests updated to include token in command payloads.
- [ ] `/pair` response documents that the token must be persisted by the client.

dev01lay2:
Closes #728

## Summary
Fixes the security vulnerability where session tokens issued at pairing were never validated on subsequent commands.

## Changes
- **Added  field** to  contract
- **Updated ** to extract session token from command input (backward compatible)
- **Added session token validation** in  for all non- commands
- **Reject requests** with missing or invalid session tokens:
  -  - when token is not provided
  -  - when token doesn't match the stored session
- **Added comprehensive test coverage** for session token validation (10 new tests)
- **Updated all existing tests** to include session tokens in their commands (282 tests passing)

## Command Format
The session token is now required for all authenticated commands:
```
<provider> <chatId> <requestId> <sessionToken> <command> [...args]
```

Example:
```
telegram 100 req-1 session-abc123 /agents
```

For backward compatibility during parsing, if the 4th field doesn't start with `session-`, it's treated as the command name (legacy format).

## Testing
- ✅ All 282 tests passing
- ✅ Type checking passed
- ✅ Added dedicated session token validation test suite
- ✅ Updated all integration, e2e, and cross-provider tests

### DIFF
--- a/gateway/src/command-routing.test.ts
+++ b/gateway/src/command-routing.test.ts
@@ -14,11 +14,12 @@ function buildDeps(): GatewayDependencies {
   };
 }
 
-function pairChat(d: GatewayDependencies, provider: "telegram" | "discord" | "feishu" = "telegram", chatId = "100"): void {
+function pairChat(d: GatewayDependencies, provider: "telegram" | "discord" | "feishu" = "telegram", chatId = "100"): string {
   if (d.daemon instanceof InMemoryDaemonClient) {
     d.daemon.registerPairCode("pair-ok");
   }
-  d.sessions.createSession({ provider, chatId });
+  const session = d.sessions.createSession({ provider, chatId });
+  return session.sessionToken;
 }
 
 describe("command routing: /pair", () => {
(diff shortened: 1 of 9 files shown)


Keith-CY (approved):
NBS: session token parsing should avoid ambiguous legacy token detection.

PR #895, opened 2026-02-15
security(gateway): restrict ARTIFACT_ROOT default and validate path

### LINKED ISSUE #865: security(gateway): ARTIFACT_ROOT defaults to cwd, allowing broad file reads
## Problem

In `gateway/src/server.ts`, the `defaultReadFile` function uses:

```ts
const artifactRoot = pathResolve(process.env.ARTIFACT_ROOT ?? process.cwd());
```

If `ARTIFACT_ROOT` is not set and the gateway is started from `/` or a broad directory, the download endpoint can serve **any file** under that tree — including source code, config files, or `.env` files.

## File

`gateway/src/server.ts` — `defaultReadFile()`

## Impact

- Download tokens are single-use and short-lived, but the daemon's `/diagnose` and `/logs` handlers write artifacts to `/tmp`, which may be under an overly permissive root
- If the gateway is started from the project root, all source and config are readable

## Suggested Fix

1. **Require** `ARTIFACT_ROOT` to be explicitly set (fail-closed if unset)
2. Or default to a narrow, dedicated directory like `/tmp/carrier-artifacts`
3. Log a warning at startup if ARTIFACT_ROOT is not configured

dev01lay2:
Closes #865

## Summary

Fixes the security issue where  defaults to current working directory (cwd), allowing potentially broad file reads.

## Changes

1. **Changed default ARTIFACT_ROOT** from  to  subdirectory - significantly reducing the default exposure
2. **Added path validation** via :
   - Blocks system root ()
   - Blocks critical system directories (, , , , , , , , )
   - Ensures all paths are resolved to absolute paths
3. **Added comprehensive tests** covering:
   - Default behavior with 
   - Reading from configured 
   - Blocking reads outside 
   - Path traversal protection
   - Validation of dangerous paths
   - Relative path resolution

## Testing

All 304 tests pass, including 11 new security tests for ARTIFACT_ROOT validation.

```bash
bun test
# 304 pass, 0 fail
```

### DIFF
--- a/gateway/src/artifact_root.test.ts
+++ b/gateway/src/artifact_root.test.ts
@@ -1,0 +1,239 @@
+import { describe, expect, test, beforeEach, afterEach } from "bun:test";
+import { createGatewayRuntime, createRuntimeDependencies } from "./server";
+import { InMemoryDaemonClient } from "./daemon/client";
+import { SessionStore } from "./session/store";
+import { DownloadTokenStore } from "./downloads/token_store";
+import { writeFile, mkdir, rm } from "node:fs/promises";
+import { join } from "node:path";
+
+function makeDeps() {
+  return {
+    daemon: new InMemoryDaemonClient(),
+    sessions: new SessionStore(),
+    downloads: new DownloadTokenStore(),
+  };
+}
+
+describe("ARTIFACT_ROOT security", () => {
+  let originalArtifactRoot: string | undefined;
+  let testDir: string;
+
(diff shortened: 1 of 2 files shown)


Akane-CN (approved):
Reviewed the latest commit and checked CI.

The ARTIFACT_ROOT boundary hardening and associated tests look good. Defaulting away from cwd significantly reduces accidental exposure.

PR #913, opened 2026-02-15
fix(daemon/api): handle /api/v1/status and /api/v1/logs agent IDs correctly

### LINKED ISSUE #909: bug(daemon/api): /api/v1/status and /api/v1/logs aliases are wired but return 400 for valid agent IDs
## Problem
`buildHTTPMux` registers `/api/v1/...` aliases for compatibility, but dynamic routes reuse handlers that trim only the unversioned prefix:
- `/api/status/` handler uses `strings.TrimPrefix(r.URL.Path, "/api/status/")`
- `/api/logs/` handler uses `strings.TrimPrefix(r.URL.Path, "/api/logs/")`

When called via `/api/v1/status/{agentId}` or `/api/v1/logs/{agentId}`, the trim does not match, so the raw value still contains slashes (for example `/api/v1/status/openclaw`). `parsePathAgentID` then rejects it as invalid and returns `400`.

## Why it matters
- Breaks advertised backward-compatibility for versioned clients on two core endpoints.
- Creates a partial migration trap: `/api/v1/agents` works while `/api/v1/status/*` and `/api/v1/logs/*` fail.
- Increases support/debug time because failures look like bad client input rather than route wiring mismatch.

- Refactor path extraction so handlers do not depend on hardcoded single-prefix trimming.
  - Option A: register dedicated handlers for `/api/...` and `/api/v1/...` with explicit prefix stripping per route.
  - Option B: centralize a helper that accepts allowed prefixes and returns the normalized `{agentId}` segment.
- Add focused regression tests for both versioned and unversioned dynamic routes:
  - `/api/status/{id}` + `/api/v1/status/{id}`
  - `/api/logs/{id}` + `/api/v1/logs/{id}`
- Keep invalid-ID behavior unchanged after normalization.

## Acceptance criteria
- [ ] `/api/v1/status/{agentId}` returns the same status code/body semantics as `/api/status/{agentId}` for valid and invalid IDs.
- [ ] `/api/v1/logs/{agentId}` returns the same status code/body semantics as `/api/logs/{agentId}`.
- [ ] New regression tests fail on current `main` and pass with the fix.
- [ ] Existing API handler tests continue to pass.

- `cd daemon && go test ./cmd/agentd -run 'TestAPI(Status|Logs).*V1|TestAPI(Status|Logs)'`

Akane-CN:
## Summary
- fix dynamic path normalization for versioned aliases on status/logs endpoints
- add a shared path-prefix helper used by both `/api/*` and `/api/v1/*` dynamic handlers
- add regression tests covering v1 status/logs for valid and invalid agent IDs

## Validation
- cd daemon && go test ./cmd/agentd -run 'TestAPI(Status|Logs|StatusAndLogs).*V1|TestAPIStatusNotFound'
- cd daemon && go test ./...

Closes #909

### DIFF
--- a/daemon/cmd/agentd/main.go
+++ b/daemon/cmd/agentd/main.go
@@ -228,7 +228,7 @@ func buildHTTPMux(svc *lifecycle.Service, ready *atomic.Bool) *http.ServeMux {
 			writeJSONError(w, http.StatusMethodNotAllowed, "method not allowed")
 			return
 		}
-		raw := strings.TrimPrefix(r.URL.Path, "/api/status/")
+		raw := trimPathByPrefixes(r.URL.Path, "/api/status/", "/api/v1/status/")
 		agentID, err := parsePathAgentID(raw)
 		if err != nil {
 			writeJSONError(w, http.StatusBadRequest, err.Error())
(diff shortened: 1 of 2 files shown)


dev01lay2 (approved):
LGTM. Clean helper function, tests cover both valid and invalid agent ID cases for the v1 aliases.

PR #917, opened 2026-02-15
fix(ci): make Kanban sync fail-open when CARRIER_PROJECT_ID is missing

Akane-CN:
## Summary
Fix Kanban sync CI false failures caused by missing `CARRIER_PROJECT_ID`.

## Change
- In `.github/workflows/carrier-kanban-automation.yml`, changed missing project-id behavior from:
  - `core.setFailed(...)`
  to
  - `core.warning(...)` + early return

## Why
`sync` was failing on unrelated PRs due to workflow configuration state, even when all core code checks were green. This makes the Kanban sync job fail-open for missing project configuration and prevents unrelated PRs from being blocked.

### DIFF
--- a/.github/workflows/carrier-kanban-automation.yml
+++ b/.github/workflows/carrier-kanban-automation.yml
@@ -22,7 +22,7 @@ jobs:
           script: |
             const projectId = process.env.CARRIER_PROJECT_ID;
             if (!projectId) {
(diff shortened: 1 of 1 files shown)


dev01lay2 (approved):
LGTM. Failing open for a missing optional env var in a non-critical sync job is the right call. Clean one-liner.

PR #963, opened 2026-02-15
fix(start): correct --systemd --user target and activation flow

### LINKED ISSUE #962: bug(scripts/start): --systemd --user writes wrong WantedBy target and behavior mismatches help text
## Problem
`scripts/start.sh` advertises `--systemd` as "Create and enable systemd service", but in non-dry-run mode it only writes the unit and runs `daemon-reload`; it does **not** execute `enable`/`start`.

Additionally, for `--systemd --user` it still writes:

```ini
[Install]
WantedBy=multi-user.target
```

which is a system target and is incorrect for user services (expected `default.target`).

## Impact / Risk
- User-level systemd units may not enable correctly due to wrong target.
- Script behavior diverges from help text, causing failed onboarding and confusing "service created but not running" outcomes.
- Increases manual recovery/support burden because users must discover/fix target and enable/start steps themselves.

## Proposed change
1. Align implementation with usage contract:
   - Either actually run `systemctl enable/start` (and possibly `--now`) in non-dry-run mode, or
   - Update help text to explicitly say "create only" and print next-step commands.
2. Generate `[Install] WantedBy=default.target` when `--user` is selected.
3. Add a small regression test (or shell assertion) that validates emitted unit target differs between system and user modes.

## Acceptance criteria
- [ ] `--systemd --user` emits `WantedBy=default.target`.
- [ ] `--systemd` path behavior matches documented help text.
- [ ] Non-dry-run flow is deterministic and clearly reports whether enable/start were executed.
- [ ] Dry-run output mirrors real-mode command sequence.

- Run `./scripts/start.sh --systemd --user` in dry-run and assert target + commands.
- Run `./scripts/start.sh --systemd` in dry-run and assert system target usage.
- Add a shell test that parses generated unit content for correct `WantedBy` by mode.

Akane-CN:
## Summary
- fix `scripts/start.sh` to emit `WantedBy=default.target` when `--systemd --user` is used
- keep `WantedBy=multi-user.target` for system-level services
- align non-dry-run behavior with help text by actually running `enable` + `start` after `daemon-reload`
- add `scripts/start_systemd_test.sh` to assert generated target and dry-run command sequence for system/user modes

## Validation
- `scripts/start_systemd_test.sh`

Closes #962

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -5,8 +5,8 @@
 # Usage:
 #   ./start.sh                    # Start daemon in foreground
 #   ./start.sh --daemon           # Start daemon in background
-#   ./start.sh --systemd          # Create and enable systemd service
-#   ./start.sh --systemd --user   # Create user-level systemd service
+#   ./start.sh --systemd          # Create, enable, and start systemd service
+#   ./start.sh --systemd --user   # Create, enable, and start user-level systemd service
 #
 # Environment Variables:
 #   OPENCLAW_PORT       - HTTP port (default: 9090)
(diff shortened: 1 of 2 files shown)


scripts/start.sh
```diff
@@ -91,9 +91,11 @@ if [ "$MODE" = "systemd" ]; then
     if [ "$SYSTEMD_USER" = "--user" ]; then
         SERVICE_DIR="$HOME/.config/systemd/user"
         SYSTEMCTL_CMD="systemctl --user"
+        WANTED_BY_TARGET="default.target"
```
dev01lay2: NBS: Consider logging which target was selected (e.g. `echo "Using target: $WANTED_BY_TARGET"`) for easier debugging when users report systemd issues.

scripts/start_systemd_test.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
dev01lay2: NBS: Could live under a `tests/` directory for consistency with other test scripts, but fine here too.

PR #975, opened 2026-02-15
test(scripts): make check-tools test hermetic

### LINKED ISSUE #972: test(scripts): make check-tools test hermetic (no ambient PATH dependency)
## Problem
`scripts/test-check-tools.sh` currently assumes the host environment already has every required dependency (`gh`, `jq`, `go`, `bun`) installed and available on `PATH`.

Because of that, the "all tools present" assertion is environment-dependent and can fail on clean/dev containers even when `scripts/check-tools.sh` itself is correct.

## Impact / Risk
- Produces false-negative test failures unrelated to repository logic.
- Reduces confidence in script-test results across different contributor/dev environments.
- Encourages skipping tests rather than trusting test outcomes.

## Proposed change
Make `scripts/test-check-tools.sh` hermetic:
- Create a temporary bin directory with stub executables for required tools used by `check-tools.sh`.
- Ensure each stub supports `--version` so the checker can run unchanged.
- Run the "success" path with `PATH` pointed to the stub bin (plus shell essentials if needed).
- Keep the "missing tools" path test (empty/nonexistent `PATH`) to verify failure behavior.

## Acceptance criteria
- [ ] `scripts/test-check-tools.sh` passes in environments that do **not** have `gh`/`jq`/`go`/`bun` installed.
- [ ] Test no longer depends on ambient host toolchain for the success-path assertion.
- [ ] Missing-tools failure-path assertion remains covered.
- [ ] Temporary test artifacts are cleaned up automatically.

## Suggested validation/tests
- Run `bash scripts/test-check-tools.sh` in:
  - a normal dev environment
  - a minimal container with required tools absent
- Verify both runs pass and report deterministic assertions.

Akane-CN:
## Summary
- make `scripts/test-check-tools.sh` hermetic for the success path by stubbing required tools (`gh`, `jq`, `go`, `bun`) in a temporary bin directory
- avoid ambient PATH dependency by invoking the checker with a controlled PATH
- keep and validate the missing-tools failure path with a nonexistent PATH
- clean up temporary artifacts via trap

## Testing
- bash scripts/test-check-tools.sh

Closes #972

### DIFF
--- a/scripts/test-check-tools.sh
+++ b/scripts/test-check-tools.sh
@@ -9,17 +9,36 @@ fail=0
 
 echo "Testing check-tools.sh..."
 
-# Test 1: With full PATH, should succeed
-if bash "$CHECK_SCRIPT" >/dev/null 2>&1; then
-  echo "  PASS: exits 0 with all tools present"
+TMP_BIN="$(mktemp -d)"
+trap 'rm -rf "$TMP_BIN"' EXIT
+
+make_stub() {
+  local tool="$1"
+  cat >"$TMP_BIN/$tool" <<'STUB'
+#!/usr/bin/env bash
+echo "stub 0.0.0"
+STUB
+  chmod +x "$TMP_BIN/$tool"
+}
+
+# Hermetic stubs for required tools used by check-tools.sh.
+make_stub "gh"
+make_stub "jq"
+make_stub "go"
+make_stub "bun"
+ln -sf "$(command -v bash)" "$TMP_BIN/bash"
+
+# Test 1: Hermetic success path with stubbed required tools.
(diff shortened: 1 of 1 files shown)


scripts/test-check-tools.sh
```diff
@@ -9,17 +9,36 @@ fail=0
...
-if bash "$CHECK_SCRIPT" >/dev/null 2>&1; then
-  echo "  PASS: exits 0 with all tools present"
+TMP_BIN="$(mktemp -d)"
+trap 'rm -rf "$TMP_BIN"' EXIT
+
+make_stub() {
+  local tool="$1"
+  cat >"$TMP_BIN/$tool" <<'STUB'
+#!/usr/bin/env bash
+echo "stub 0.0.0"
```
dev01lay2: NBS: Minor — `make_stub` could use `printf` instead of heredoc for the one-liner, but this is perfectly readable as-is.

PR #1014, opened 2026-02-15
test(api): lock malformed /api/v1/agents path contract with focused parser-route tests

### LINKED ISSUE #1004: test(api): lock malformed /api/v1/agents path contract with focused parser-route regression suite
## Problem
`cmd/agentd` currently has strong parity coverage for **happy-path and method checks** between legacy routes and `/api/v1/agents/{id}/{action}`, but it does not lock the malformed-path contract for v1 route parsing.

`parseAgentActionPath` returns a generic `ok=false` for many invalid forms (double slashes, invalid URL-escape, path traversal-ish tokens, empty action), and the mux currently falls back to `http.NotFound`. That behavior can silently drift because malformed-path expectations are not asserted in focused tests.

## Why it matters
- Clients integrating against `/api/v1/agents/*` need deterministic error behavior for bad input.
- Security hardening depends on rejecting malformed IDs/path patterns consistently.
- Refactors in router/parser logic are high-risk without regression tests for these failure paths.

1. Add focused table-driven tests for malformed v1 paths (separate from broad lifecycle parity tests), covering at least:
   - `//` in path segments
   - invalid percent-encoding
   - encoded slash / backslash attempts
   - parent-directory token attempts (`..`)
   - missing/blank action segment
2. Assert **status + response contract** per case (e.g., 404 vs JSON 4xx) and document the intended contract in test names/comments.
3. Refactor route parsing/error mapping if needed so parser outcomes are explicit and easier to maintain (e.g., distinguish "not matched" vs "matched but invalid").
4. Keep legacy/v1 parity tests, but remove malformed-path assertions from broad parity flows in favor of the new focused suite.

## Acceptance criteria
- [ ] New table-driven malformed-path tests exist for `/api/v1/agents/{id}/{action}`.
- [ ] Tests cover invalid encoding, traversal attempts, empty segments, and double-slash forms.
- [ ] Current and intended HTTP/error-body contract is explicitly asserted.
- [ ] Router/parser code is refactored only as needed to support clear, stable error mapping.
- [ ] `go test` passes for `daemon/cmd/agentd`.

```bash
cd daemon

go test ./cmd/agentd -run 'Test(V1AgentRoute|ParseAgentActionPath|LegacyV1Parity)' -count=1

go test ./cmd/agentd -count=1
```

dev01lay2:
Add `agents_path_test.go` with three focused test functions:

- **TestParseAgentActionPath** (21 cases): valid paths, missing agent ID, missing action, extra path segments, empty segments (double slashes), special characters (encoded slashes, backslashes, traversal, unicode, spaces), wrong prefixes
- **TestAgentsPathHTTPRouting** (13 cases): HTTP-level verification through the full mux — valid routes, bare trailing slash, ID without action, extra segments, unknown actions, empty ID segments, traversal attempts, encoded slashes, non-existent agents, and the `/api/v1/agents/status` special case
- **TestParsePathAgentIDMalformed** (11 cases): direct parser helper coverage for empty input, path separators, dot-dot traversal, encoded slashes, leading dots/hyphens, spaces, and invalid percent encoding

Closes #1004

### DIFF
--- a/daemon/cmd/agentd/agents_path_test.go
+++ b/daemon/cmd/agentd/agents_path_test.go
@@ -1,0 +1,162 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"carrier/daemon/internal/baseagent"
+	"carrier/daemon/internal/catalog"
+	"carrier/daemon/internal/lifecycle"
+)
+
+func TestParseAgentActionPath(t *testing.T) {
+	tests := []struct {
+		name       string
+		path       string
+		wantID     string
+		wantAction string
+		wantOK     bool
+	}{
+		// Valid cases
+		{name: "valid", path: "/api/v1/agents/myagent/status", wantID: "myagent", wantAction: "status", wantOK: true},
+		{name: "dots_hyphens_underscores", path: "/api/v1/agents/my.agent_1-2/logs", wantID: "my.agent_1-2", wantAction: "logs", wantOK: true},
+
+		// Missing agent ID
+		{name: "no_id_no_action", path: "/api/v1/agents/", wantOK: false},
+		{name: "only_action", path: "/api/v1/agents//status", wantOK: false},
+
+		// Missing action
+		{name: "id_no_action", path: "/api/v1/agents/myagent", wantOK: false},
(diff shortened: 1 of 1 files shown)


Keith-CY (approved):
LGTM

PR #1017, opened 2026-02-15
fix(daemon): rotate agent log files to prevent unbounded growth

### LINKED ISSUE #687: bug(daemon): agent process log files grow unbounded on disk
## Problem

`ProcessManager.Start()` opens a per-agent log file at `{logDir}/{agentID}.log` with `os.O_APPEND` and pipes both stdout and stderr into it. There is no log rotation, size cap, or truncation mechanism.

A long-running agent producing verbose output can fill the disk, causing cascading failures (state file writes fail, new processes can't start, etc.).

Note: The in-memory `s.logs[agentID]` ring buffer in `lifecycle/helpers.go` is capped by `logLimit`, but the **on-disk log file** written by ProcessManager has no such limit.

## Location

- `daemon/internal/lifecycle/process.go:Start()` — line opening logFile with O_APPEND

## Suggested Fix

Either:
1. Rotate the log file when it exceeds a configurable size (e.g., 10MB), keeping 1-2 rotated copies
2. Use a size-capped writer wrapper that truncates from the beginning
3. Periodically truncate in the monitoring goroutine

dev01lay2:
## Problem
Agent process logs in the daemon write to `{logDir}/{agentID}.log` using `os.O_APPEND` but never rotate or truncate. On long-running agents, this file grows without bound.

## Fix
Added `rotateLogFile()` which is called before each process start. When the log file exceeds **10 MB**, it is renamed to `.log.1` (keeping one generation of backup) and a fresh log file is created.

- Simple, zero-dependency approach — no external log rotation libraries
- One backup file kept (`.log.1`) so recent logs aren't lost
- Rotation happens at process start, not during writes, so no locking complexity

## Tests
Added `rotate_test.go` with 5 test cases covering:
- Missing file (no-op)
- File under threshold (no-op)
- File over threshold (rotated)
- Previous backup overwritten
- Integration: `ProcessManager.Start` rotates oversized log

Closes #687

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   e2e-deploy:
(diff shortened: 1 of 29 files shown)


Keith-CY (approved):
LGTM

PR #1019, opened 2026-02-15
fix(lifecycle): replace sequential handoff ID counter with crypto/rand (#729)

### LINKED ISSUE #729: bug(daemon): handoff ID generator resets to zero on restart — collision risk for diagnosis handoffs
## Problem

`Service.idCounter` is initialized to `0` on every `NewService()` call. The default `idGenerator` produces IDs like `handoff-1`, `handoff-2`, etc. After a daemon restart, the counter resets, producing **duplicate handoff IDs** that collide with previously issued handoffs.

**File:** `daemon/internal/lifecycle/service.go`
```go
svc.idGenerator = func(prefix string) string {
    next := atomic.AddUint64(&svc.idCounter, 1)
    return fmt.Sprintf("%s-%d", prefix, next)
}
```

## Impact
- Diagnosis handoffs created after restart overwrite entries in `s.handoffs` map with the same key.
- If a gateway client caches a handoff ID, it may resolve to a completely different handoff after daemon restart.
- Audit trail becomes unreliable — handoff references are not globally unique.

## Proposed Fix
Use a UUID or timestamp-seeded counter for handoff IDs:
```go
svc.idGenerator = func(prefix string) string {
    return fmt.Sprintf("%s-%s", prefix, uuid.New().String())
}
```
Or persist the counter in the state file and restore it on startup.

## Acceptance Criteria
- [ ] Handoff IDs are unique across daemon restarts.
- [ ] Existing tests pass with the new ID scheme.
- [ ] Audit log entries referencing handoff IDs remain unambiguous.

dev01lay2:
## Problem
The handoff ID generator used an atomic counter starting at 0 (`atomic.AddUint64`), which reset on every daemon restart, creating collision risk for handoff IDs.

## Fix
- Replace sequential counter with `crypto/rand`-based 128-bit hex IDs (e.g. `handoff-a1b2c3d4...`)
- Remove unused `idCounter` field and `sync/atomic` import
- Add test verifying ID uniqueness across two independent service instances (simulating restart)

All tests pass. Fixes #729

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   e2e-deploy:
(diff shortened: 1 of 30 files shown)


Keith-CY (approved):
LGTM

PR #1022, opened 2026-02-16
fix(daemon): clean up ProcessManager entries and FDs after natural process exit

### LINKED ISSUE #655: bug(daemon): ProcessManager leaks file descriptors and stale entries after natural process exit
## Problem
`ProcessManager` only closes `info.logFile` and removes `pm.processes[agentID]` in `Stop()`. When an agent process exits on its own:

- `cmd.Wait()` completes (`process.go:90-94`), but
- no cleanup runs for that path, so the process entry remains in `pm.processes`, and
- the corresponding log file handle remains open until a later explicit `Stop()` (which may never happen).

Relevant code:
- `daemon/internal/lifecycle/process.go:90-94` (`Wait` goroutine records completion only)
- `daemon/internal/lifecycle/process.go:133-139` (cleanup only in `Stop`)
- `daemon/internal/lifecycle/process.go:50-51` (stale map entry deleted on next `Start`, but without closing prior `logFile`)

## Impact / Risk
- FD/resource leak over time when agents crash/exit without explicit stop.
- Long-running daemon can accumulate stale process metadata and open file handles.
- In high churn scenarios this can degrade reliability and eventually hit open-file limits.

## Proposed change
Add a unified cleanup path for process completion that runs for **both** explicit stop and natural exit:

- Introduce an internal `finalizeProcess(agentID, info)` helper that:
  - closes `logFile` exactly once,
  - removes `pm.processes[agentID]` if it still points to `info`.
- Invoke it from the `cmd.Wait()` goroutine after signaling `done`.
- Keep `Stop()` using `sync.Once` semantics and call the same finalize helper.
- Add tests to verify stale-entry + FD cleanup on natural exit.

## Acceptance criteria
- [ ] After a naturally exiting process, `pm.processes` no longer contains that `agentID`.
- [ ] Log file descriptor is closed even when `Stop()` is never called.
- [ ] Restarting same `agentID` after natural exit does not depend on stale-entry deletion side effects.
- [ ] `go test ./daemon/internal/lifecycle -run ProcessManager -count=1` passes.

## Suggested validation/tests
- Add a test that starts short-lived command (`sleep 0.1`), waits for completion, then asserts internal map cleanup.
- Add a test that performs repeated start/exit cycles and checks no monotonic growth of open FDs (Linux: `/proc/self/fd` count heuristic).
- Run full daemon tests: `cd daemon && go test ./...`.

dev01lay2:
## Summary

Fixes #655 — ProcessManager leaks file descriptors and stale entries after natural process exit.

## Changes

- Add `finalizeProcess(agentID, info)` helper that:
  - Closes `logFile` exactly once via `sync.Once`
  - Removes `pm.processes[agentID]` if it still points to the same `processInfo`
- Call `finalizeProcess` from the `cmd.Wait()` goroutine after `close(info.done)` — handles natural exit cleanup
- Replace direct cleanup in `Stop()` with `finalizeProcess` — both paths now use the same unified cleanup
- Fix `Start()` stale-entry cleanup to close the old log file (previously leaked)

## Tests Added

- `TestProcessManager_NaturalExitCleansUpEntry` — verifies map entry removal after natural exit
- `TestProcessManager_NaturalExitClosesLogFile` — verifies FD cleanup without explicit Stop
- `TestProcessManager_RestartAfterNaturalExit` — verifies restart works cleanly after natural exit

### DIFF
--- a/daemon/internal/lifecycle/process.go
+++ b/daemon/internal/lifecycle/process.go
@@ -19,14 +19,15 @@ type ProcessManager struct {
 }
 
 type processInfo struct {
-	cmd      *exec.Cmd
-	pid      int
-	agentID  string
-	logFile  *os.File
-	done     chan struct{}
-	waitErr  error
-	exitCode *int
-	stopping bool
+	cmd          *exec.Cmd
+	pid          int
+	agentID      string
+	logFile      *os.File
+	done         chan struct{}
+	waitErr      error
+	exitCode     *int
+	stopping     bool
+	closeLogOnce sync.Once
 }
 
 // NewProcessManager creates a new process manager.
@@ -49,7 +50,10 @@ func (pm *ProcessManager) Start(agentID string, command string, args []string) (
 		if pm.isProcessAlive(info) {
 			return 0, fmt.Errorf("agent %s already running with PID %d", agentID, info.pid)
 		}
-		// Clean up stale entry
+		// Clean up stale entry (close log file if not already closed)
+		info.closeLogOnce.Do(func() {
+			info.logFile.Close()
+		})
 		delete(pm.processes, agentID)
 	}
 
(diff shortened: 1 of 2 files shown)


Keith-CY (changes requested):
Request changes: cleanup now removes process state before callers can query exit status on natural exit.

dev01lay2:
Addressed in 9eb3a8e: the wait goroutine no longer calls `finalizeProcess` — it only closes the log file via `closeLogOnce`. The process entry is now retained in the map after natural exit so callers can still read `exitCode` via `GetExitCode()`. Removal is deferred to either `Stop()` or the next `Start()` for the same agentID.

Ready for re-review.

Keith-CY (approved):

PR #1023, opened 2026-02-16
fix(daemon): cap agent process log files with rotation

### LINKED ISSUE #687: bug(daemon): agent process log files grow unbounded on disk

dev01lay2:
## Summary

Fixes #687 — agent process log files grow unbounded on disk.

## Changes

- **New `cappedFileWriter`** (`capped_writer.go`): An `io.Writer` wrapper that rotates the underlying log file when it exceeds 10 MB, keeping one rotated copy (`.log.1`). Thread-safe via mutex.
- **Updated `ProcessManager.Start()`** (`process.go`): Replaced raw `os.OpenFile` with `newCappedFileWriter`, so all per-agent log files are now size-capped.
- **Added `closeLogOnce sync.Once`** to `processInfo` struct to fix the pre-existing usage of `sync.Once` for log file cleanup.
- **Tests** (`capped_writer_test.go`): Covers rotation trigger, existing file accounting, and default max bytes.

## Behavior

| Before | After |
|--------|-------|
| Log files grow without limit | Rotated at 10 MB, max ~20 MB per agent (current + 1 rotated) |

The in-memory ring buffer (`logLimit`) was already capped; this fix addresses the on-disk file.

### DIFF
--- a/daemon/internal/lifecycle/capped_writer.go
+++ b/daemon/internal/lifecycle/capped_writer.go
@@ -1,0 +1,92 @@
+package lifecycle
+
+import (
+	"fmt"
+	"os"
+	"sync"
+)
+
+const (
+	// defaultMaxLogBytes is the default maximum size of a log file before rotation (10 MB).
+	defaultMaxLogBytes int64 = 10 * 1024 * 1024
+)
+
+// cappedFileWriter wraps an os.File and rotates it when the written bytes
+// exceed maxBytes. It keeps one rotated copy (e.g. agent.log.1).
+// All methods are safe for concurrent use.
+type cappedFileWriter struct {
+	mu       sync.Mutex
+	file     *os.File
+	path     string
+	written  int64
+	maxBytes int64
+}
+
+// newCappedFileWriter opens (or creates) the file at path in append mode and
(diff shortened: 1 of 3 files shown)


Keith-CY (approved):
LGTM

PR #1025, opened 2026-02-16
fix(daemon): close Stop/monitorProcess race with RuntimeStateStopping

### LINKED ISSUE #659: bug(daemon): Stop() races with monitorProcess(), causing false crash-loop triggers
## Summary

When `Stop()` is called on a running agent, there is a race condition between `Stop()` and the `monitorProcess()` goroutine that can cause a clean stop to be misinterpreted as a crash, potentially triggering false crash-loop detection.

## Root Cause

**File:** `daemon/internal/lifecycle/start_stop.go` + `daemon/internal/lifecycle/process.go`

The race window:

1. `Stop()` calls `pm.Stop(agentID)` which sends SIGTERM and waits on `info.done`
2. Process exits → the Start goroutine sends to `info.done` (buffered 1) and closes the channel
3. `pm.Stop()` reads from `info.done`, closes the log file, and deletes the entry from `pm.processes`
4. `Stop()` hasn't yet set `state.Runtime = RuntimeStateStopped`
5. Meanwhile, `monitorProcess()` calls `pm.Wait(agentID)`:
   - **If the map entry hasn't been deleted yet:** reads nil from the already-closed channel
   - Then checks `state.Runtime == RuntimeStateRunning` — **still true** because Stop hasn't updated state yet
   - Sets state to `RuntimeStateCrashing` and calls `recordRestart()`
6. Stop() then sets `RuntimeStateStopped`, overwriting the crash state

**Net effect:** A clean stop gets recorded as a restart in the crash-loop counter. After 3 clean stop/start cycles within the crash-loop window, the agent enters false crash-loop cooldown and refuses to start.

## Suggested Fix

Coordinate between `Stop()` and `monitorProcess()`:
- Option A: Set a "stopping" flag before calling `pm.Stop()`, and have `monitorProcess()` check it before marking as crashed
- Option B: Cancel the `monitorProcess` goroutine (via context) when `Stop()` is invoked
- Option C: Set `state.Runtime` to an intermediate state (e.g. `RuntimeStateStopping`) before calling `pm.Stop()`, and have `monitorProcess()` only react to `RuntimeStateRunning`

dev01lay2:
## Summary

Fixes #659 — Stop() races with monitorProcess(), causing false crash-loop triggers.

## Root Cause

Stop() used two separate lock acquisitions: `getManifestAndState()` (RLock) followed by a write-Lock to set `RuntimeStateStopped`. In the gap between them, `monitorProcess()` could acquire the lock, observe `RuntimeStateRunning`, and call `recordRestart()` — recording a false crash.

## Fix

1. **New `RuntimeStateStopping` state** — intermediate state between Running and Stopped
2. **Atomic check-and-transition** — Stop() checks preconditions and sets `RuntimeStateStopping` in a single Lock acquisition, eliminating the race window
3. **monitorProcess() unchanged** — it only reacts to `RuntimeStateRunning`, so it naturally ignores processes being intentionally stopped
4. **Final transition** — after successful process termination, Stop() sets `RuntimeStateStopped`

## Changes

- `types.go`: Add `RuntimeStateStopping` constant
- `start_stop.go`: Rewrite Stop() to atomically transition to Stopping before killing the process; set Stopped after success

### DIFF
--- a/daemon/internal/lifecycle/start_stop.go
+++ b/daemon/internal/lifecycle/start_stop.go
@@ -116,19 +116,22 @@ func (s *Service) Start(ctx context.Context, agentID string) error {
 }
 
 func (s *Service) Stop(ctx context.Context, agentID string) error {
-	_, state, err := s.getManifestAndState(agentID)
-	if err != nil {
-		return err
+	// Atomically check preconditions and transition to RuntimeStateStopping in
+	// a single lock acquisition.  This closes the race window where
+	// monitorProcess() could observe RuntimeStateRunning between the old
+	// getManifestAndState (RLock) and the subsequent write-Lock, recording a
+	// false crash-loop restart.
+	s.mu.Lock()
+	if _, ok := s.manifests[agentID]; !ok {
+		s.mu.Unlock()
+		return ErrAgentNotFound
 	}
-	if state.Runtime == RuntimeStateStopped {
+	state := s.states[agentID]
+	if state.Runtime == RuntimeStateStopped || state.Runtime == RuntimeStateStopping {
+		s.mu.Unlock()
 		return ErrAlreadyStopped
 	}
-
-	// Mark as stopped before killing the process so that monitorProcess
-	// (which races on the same lock) won't overwrite the state to Crashing.
-	s.mu.Lock()
-	state = s.states[agentID]
-	state.Runtime = RuntimeStateStopped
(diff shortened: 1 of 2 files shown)


Akane-CN (approved):
NBS: RuntimeStateStopping race fix is scoped and sensible, with passing checks. Approved.

Keith-CY (approved):

PR #1026, opened 2026-02-16
fix(daemon): stop agents concurrently during shutdown to respect shutdownTimeout

### LINKED ISSUE #686: bug(daemon): sequential agent shutdown can exceed shutdownTimeout with multiple agents
## Problem

`stopAllAgents()` in `daemon/cmd/agentd/main.go` stops agents sequentially. Each `svc.Stop()` call triggers `ProcessManager.Stop()`, which sends SIGTERM and waits **up to 10 seconds** before sending SIGKILL.

With the `shutdownTimeout` set to 30 seconds, just 4 unresponsive agents would blow past the deadline (4 × 10s = 40s > 30s), causing `context.DeadlineExceeded` and a `log.Fatalf` — an unclean shutdown.

## Location

- `daemon/cmd/agentd/main.go`: `stopAllAgents()` — sequential loop
- `daemon/internal/lifecycle/process.go`: `Stop()` — 10s grace period per agent

## Suggested Fix

Stop agents concurrently with a shared context derived from the shutdown timeout:

```go
func stopAllAgents(svc *lifecycle.Service, timeout time.Duration) error {
    ctx, cancel := context.WithTimeout(context.Background(), timeout)
    defer cancel()
    agents := svc.ListAgents()
    var wg sync.WaitGroup
    // ... stop concurrently, respect ctx ...
}
```

Or pass the remaining shutdown budget to each `Stop()` call.

dev01lay2:
## Problem

`stopAllAgents()` stopped agents sequentially. Each `ProcessManager.Stop()` blocks up to 10s for graceful termination. With N unresponsive agents this took N×10s, exceeding the 30s `shutdownTimeout` (e.g. 4 agents × 10s = 40s > 30s), causing `context.DeadlineExceeded` and `log.Fatalf`.

## Fix

Stop agents concurrently using `sync.WaitGroup`. The shutdown context is passed through so all stops share the same deadline. Total shutdown time is now bounded by `max(single agent stop time)` rather than `sum(all agent stop times)`.

Fixes #686

### DIFF
--- a/daemon/cmd/agentd/main.go
+++ b/daemon/cmd/agentd/main.go
@@ -16,6 +16,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
@@ -623,7 +624,7 @@ func shutdownAgents(svc *lifecycle.Service, timeout time.Duration) error {
 
 	done := make(chan error, 1)
 	go func() {
-		done <- stopAllAgents(svc)
+		done <- stopAllAgents(ctx, svc)
 	}()
 
 	select {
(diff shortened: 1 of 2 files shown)


Akane-CN (approved):
NBS: Concurrent shutdown handling is targeted and improves timeout behavior without broad side effects. CI green. Approved.

Keith-CY (approved):

PR #1030, opened 2026-02-16
feat: implement /status canonical fields (health/runtime/version/ports/uptime/restart_count)

### LINKED ISSUE #808: [Micro][T016][#30] Implement `/status` canonical fields (`health/runtime/version/ports/uptime/restart_count`).
## Micro Task
Implement `/status` canonical fields (`health/runtime/version/ports/uptime/restart_count`).

## Metadata
- Parent Decomposition: #30
- Roll-up Tracker: #119
- Task Code: `T016`
- Owner: @dev01lay2
- Size: `S`
- WIP Policy: max 2 active tasks per owner

## Dependencies
- Human: `T011`
- depends-on: #803

## DoR (Definition of Ready)
- [ ] Exact file/function scope is listed in first implementation comment.
- [ ] Expected API/behavior contract is stated (or linked).
- [ ] Target tests/specs are identified before code change.
- [ ] All `depends-on` issues are closed.
- [ ] For API/behavior tasks: spec-first required = `yes`.

## DoD (Definition of Done)
- [ ] Exactly one functional change aligned to this micro task.
- [ ] Focused tests/spec updates are included and passing.
- [ ] Docs are updated when API/contract behavior changes.
- [ ] CI is green on the PR.
- [ ] PR body includes evidence links and references parent/tracker issues.

## Parent Closure Contract
- Parent issue closes only after all required micro tasks are closed and evidence links are present.

dev01lay2:
## Summary

Implements the `/status` canonical fields as specified in #808:

- **health**: agent health state (healthy/unhealthy/unknown)
- **runtime**: runtime state (running/stopped)
- **version**: agent version string
- **ports**: list of ports from manifest (populated on start, cleared on stop)
- **uptime**: human-readable uptime computed from startedAt timestamp
- **restart_count**: number of restarts (incremented on each start after the first)

### Daemon (Go)
- Added `Ports`, `StartedAt`, `RestartCount` fields to `AgentState` struct
- Populated ports from manifest on agent start, cleared on stop
- Track `StartedAt` timestamp and increment `RestartCount` on restarts

### Gateway (TypeScript)
- Added matching fields to `DaemonAgentState` type
- Updated `InMemoryDaemonClient` to track new fields
- Updated `/status` response format to display all canonical fields
- Added `formatUptime()` helper for human-readable uptime

## Testing
- All 337 existing gateway tests pass

Closes #808


.gitignore
```diff
@@ -6,3 +6,5 @@ daemon/coverage.out
 .worktrees/
 agentd
 !daemon/cmd/agentd/
+carrier
```
Keith-CY: Please avoid adding `carrier` here. This pattern is broad enough to hide a top-level path named `carrier` and is unrelated to a specific build artifact; if you need to ignore temporary artifacts, use a narrower path (or a local ignore) and document it.

Keith-CY (changes requested):
Blocking: there is a broad repo-level ignore side-effect in this change that can hide valid tracked paths and appears tied to local cleanup only.

dev01lay2:
Addressed both blockers in eecc288:

1. **RestartCount**: Moved `state.StartedAt = &now` after the nil check so `RestartCount` only increments on actual restarts, not the first start.
2. **Ports nil→[]**: `Stop` now sets `state.Ports = []int{}` instead of `nil`, ensuring JSON serializes as `[]` matching the `number[]` API schema.
3. **TS InMemoryDaemonClient**: Changed `restartCount` logic from `state.startedAt ? count+1 : 0` to `state.startedAt ? count+1 : count` so the count isn't reset after stop/start cycles.

All tests pass (Go lifecycle + 337 gateway). @Linn-San ready for re-review.

dev01lay2:
Thanks for catching this @Keith-CY. Fixed in 1e87601 — both `.gitignore` entries are now prefixed with `/` so they only match at the repo root and won't hide valid tracked paths in subdirectories.

dev01lay2:
All blocking issues from previous reviews have been addressed in the latest push (fa5a6c1):

1. **RestartCount on first start** — Fixed: now only increments when `StartedAt != nil` (i.e., previously started), then sets `StartedAt`.
2. **Ports nil on stop** — Fixed: now sets `Ports = []int{}` instead of `nil`.
3. **.gitignore side-effect** — Removed from this PR (no longer in diff).

CI is green. Requesting re-review. 🙏

Linn-San (approved):
No blocking findings on this pass.

I re-reviewed the latest commit range and both prior critical paths are addressed: restart counter logic now checks prior start state before update, and in-memory restartCount no longer resets across stop/start cycles. Existing behavior looks consistent with the intended status enhancements.

Approve.

PR #1045, opened 2026-02-17
security(gateway): require API token for /command on exposed hosts

### LINKED ISSUE #631: security(gateway): /command endpoint has no authentication — any network client can execute commands
## Problem

The gateway's `/command` HTTP endpoint accepts requests from **any client** without authentication. While the daemon has bearer-token auth on `/api/` routes, the gateway — which is the public-facing entry point — has none.

**File:** `gateway/src/server.ts` (lines 72–92, the `/command` route handler)

Any host that can reach the gateway port (default 8787) can:
- Pair sessions (`/pair`)
- Install/start/stop/upgrade agents
- Retrieve logs and diagnose artifacts
- Create remote diagnosis handoffs

## Impact

If the gateway is exposed on a non-loopback interface (or via port forwarding), this is a full compromise of the agent management plane. Even on loopback, any local process can issue commands.

## Suggested Fix

1. Add a `CARRIER_GATEWAY_API_TOKEN` env var (or reuse the daemon token).
2. Add bearer-auth middleware to the gateway's `/command` endpoint (similar to the daemon's `bearerAuthMiddleware`).
3. Exempt `/healthz` from auth (like the daemon does).
4. Refuse non-loopback binding without a token (like the daemon's startup check in `main.go:79-83`).

Keith-CY:
## Summary
- add gateway-level bearer auth enforcement for `POST /command` when `CARRIER_GATEWAY_API_TOKEN` is set
- compare provided/expected gateway tokens with constant-time `timingSafeEqual`
- keep `/healthz` unauthenticated
- require `CARRIER_GATEWAY_API_TOKEN` when binding the gateway to non-loopback hosts
- preserve existing session-token command auth and keep backward compatibility for Authorization-based session token usage when gateway API token is not configured
- document new auth behavior and error codes in README and command contract docs

## Why
Issue #631 reports that `/command` lacks an independent gateway API auth layer and allows non-loopback exposure without token protection.

## Test Commands and Evidence
- `cd gateway && bun run check`
- `cd gateway && bun test src/server.test.ts --test-name-pattern "command authentication|port resolution fallback behavior"`
  - Evidence: `24 pass, 0 fail` for targeted auth/binding coverage, including new tests for:
    - `E_GATEWAY_AUTH_REQUIRED`
    - `E_GATEWAY_AUTH_INVALID`
    - non-loopback bind guard with/without token

Closes #631

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Optional local flow from repo root:
   - `GET /healthz`
 - Command ingress route:
   - `POST /command` with JSON body `{ "input": "<provider> <chat_id> <request_id> <command> [...args]" }`
+  - If `CARRIER_GATEWAY_API_TOKEN` is set, send `Authorization: Bearer <gateway_api_token>`
 - Artifact download route:
   - `GET /downloads/<token>/<filename>`
 
@@ -146,6 +147,7 @@ Runtime environment variables:
 - `CARRIER_DAEMON_BASE_URL` (default: `http://127.0.0.1:9090`)
 - `CARRIER_GATEWAY_HOST` (default: `127.0.0.1`)
 - `CARRIER_GATEWAY_PORT` (default: `8787`)
+- `CARRIER_GATEWAY_API_TOKEN` (optional on loopback; required for non-loopback bind)
 
 #### Merge Queue
 - This repository supports GitHub Merge Queue for `main`.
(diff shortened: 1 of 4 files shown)


dev01lay2 (approved):
Solid security addition. Timing-safe comparison, loopback exemption, proper error codes, comprehensive test coverage including healthz bypass. CI lint + gateway green; daemon tests still pending but the changes are gateway-scoped.

One minor note:

PR #1046, opened 2026-02-17
security(daemon): hash-before-compare for bearer auth

### LINKED ISSUE #587: security(daemon): bearerAuthMiddleware uses non-constant-time string comparison
## Problem

In `daemon/cmd/agentd/main.go`, the bearer token comparison uses plain string equality:

```go
if auth != "Bearer "+token {
```

This is vulnerable to timing side-channel attacks. An attacker making repeated requests can measure response times to incrementally guess the token character by character.

## File
`daemon/cmd/agentd/main.go` — `bearerAuthMiddleware` function

## Suggested Fix

Use `crypto/subtle.ConstantTimeCompare`:

```go
import "crypto/subtle"

expected := "Bearer " + token
if subtle.ConstantTimeCompare([]byte(auth), []byte(expected)) != 1 {
    // reject
}
```

## Impact
P0 — The API token is the sole auth mechanism for non-loopback deployments. A timing attack could recover the full token.

Keith-CY:
## Summary
- harden daemon bearer auth comparison by hashing both `Authorization` and expected bearer token before `subtle.ConstantTimeCompare`
- keep `/healthz` and `/readyz` exemption behavior unchanged
- extend middleware boundary tests with explicit length-mismatch token case

## Why
Issue #587 calls out timing-side-channel risk around token comparisons. This keeps constant-time verification while avoiding early length-based comparison differences on raw strings.

## Test Commands and Evidence
- `cd daemon && CGO_ENABLED=0 go test ./cmd/agentd -run TestBearerAuthMiddlewareBoundaryCases`
  - Evidence: `ok   carrier/daemon/cmd/agentd`
- `cd daemon && CGO_ENABLED=0 go test ./cmd/agentd`
  - Evidence: `ok   carrier/daemon/cmd/agentd`

## Notes
- Local default `go test` (without `CGO_ENABLED=0`) hits a macOS linker/runtime `LC_UUID` issue in this environment; test suite passes with `CGO_ENABLED=0`.

Closes #587

### DIFF
--- a/daemon/cmd/agentd/main.go
+++ b/daemon/cmd/agentd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/subtle"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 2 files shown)


Akane-CN (approved):
NBS: Hash-before-compare auth hardening is a good security improvement with green CI. Approved.

Issue #636, opened 2026-02-14
bug(gateway): SessionStore.pair does not verify code against daemon pairing API


dev01lay2:
## Problem

When a user runs `/pair <code>`, the gateway's `handleCommand` calls `deps.sessions.pair()` which only checks its **local in-memory** `pairCodes` map. It never calls the daemon's `/api/v1/pairing/verify-consume` endpoint.

**File:** `gateway/src/index.ts` (lines 67–80) and `gateway/src/session/store.ts`

This means the gateway and daemon have **completely independent pairing stores**. A code issued by the daemon (`POST /api/v1/pairing/codes`) will never be recognized by the gateway, and vice versa.

## Impact

- The pairing flow described in docs/README cannot work end-to-end: the daemon issues a code on startup, but the gateway doesn't know about it
- This is likely the root cause of #625 (PAIR_CODE not printed) and #599 (gateway doesn't verify against daemon)

## Note

Issue #599 exists with a similar title but focuses on the architectural gap. This issue is more specific about the code path: `SessionStore.pair()` should call the daemon's verify-consume API, or the gateway should delegate code issuance to the daemon.

## Suggested Fix

In `handleCommand` for `/pair`, call `deps.daemon.verifyAndConsumePairCode(code)` before creating the session. Add a `verifyAndConsumePairCode` method to `DaemonClient`.

Issue #618, opened 2026-02-14
bug(gateway): /logs download token references file path never written to disk


dev01lay2:
## Problem

In `gateway/src/index.ts`, the `/logs` command handler creates a download token for a file that doesn't exist:

```typescript
const token = deps.downloads.issue(`/tmp/${agentId}-logs-${cmd.requestId}.txt`);
response.downloadUrl = deps.downloads.toDownloadURL(token);
```

The log lines are fetched from the daemon via `deps.daemon.getLogs()`, but they are **never written to the path** `/tmp/{agentId}-logs-{requestId}.txt`. When a user follows the download URL, `defaultReadFile` in `server.ts` will call `Bun.file(fileRef).exists()` which returns false, resulting in a 404 error.

## Location

- `gateway/src/index.ts` — `/logs` case in `handleCommand`
- `gateway/src/server.ts` — `defaultReadFile`

## Impact

- Download URLs for large log outputs are always broken (404)
- Users see "artifact file was not found" when clicking download links

## Suggested fix

Either:
1. **Write the log content to disk** before issuing the download token, or
2. **Return log lines inline** in the JSON response and remove the dead download path, or
3. **Use an in-memory blob store** that `readFile` can resolve without filesystem access

dev01lay2:
Reopening — was accidentally closed by automated cleanup.

Issue #617, opened 2026-02-14
bug(gateway): SessionStore.sessions map grows unbounded — no TTL or eviction


dev01lay2:
## Problem

`SessionStore` in `gateway/src/session/store.ts` has a `cleanup()` method that only removes expired **pairing codes**, but paired sessions stored in `this.sessions` are never evicted. Every successful `/pair` permanently adds an entry.

Over time (days/weeks of uptime), the sessions map grows without bound, consuming memory proportional to the number of unique provider:chatId pairs that have ever paired.

## Location

- `gateway/src/session/store.ts` — `SessionStore` class
- `this.sessions: Map<string, SessionRecord>` — only grows, never shrinks

## Impact

- Memory leak proportional to unique paired sessions
- No way to revoke/expire old sessions (security concern — a compromised session token is valid forever)

## Suggested fix

1. Add a `lastSeenAt`-based TTL (e.g., 30 days of inactivity → evict)
2. Add a `cleanupSessions()` method similar to existing `cleanup()` for pair codes
3. Wire it into the same periodic cleanup interval
4. Consider adding a `/unpair` or `/revoke` command for manual session removal

Issue #598, opened 2026-02-14
bug(daemon/catalog): unsupported GOOS/GOARCH falls back to linux/amd64 checksum


Akane-CN:
## Problem
`daemon/internal/catalog/manifests.go` `getPinnedChecksum()` defaults to `openclawChecksumLinuxX86` for unknown platforms:

```go
default:
    return openclawChecksumLinuxX86
```

For unsupported `GOOS/GOARCH` combinations, this returns a checksum for a different artifact instead of signaling unsupported platform.

## Impact / Risk
- Platform mismatch can produce misleading install behavior on unsupported targets.
- Error semantics become confusing (`checksum mismatch`/installer failure instead of explicit unsupported-platform failure).
- Harder triage for users and maintainers; weakens fail-closed clarity in installer path.

## Proposed change
- Change default branch to return empty string (or explicit unsupported error signal) for unknown platforms.
- In `getInstallCommand`, emit a deterministic fatal message for unsupported platform before executing installer.
- Add tests for supported mappings plus unsupported fallback behavior.

## Acceptance criteria
- [ ] Unknown `GOOS/GOARCH` no longer reuses linux/amd64 checksum.
- [ ] Install command fails fast with clear unsupported-platform error.
- [ ] Supported platform mappings continue to return the correct pinned checksum.
- [ ] Unit tests cover both supported and unsupported platform branches.

## Suggested validation/tests
- Add table-driven tests around checksum selection/mapping.
- Validate generated install command path for unsupported platforms (expected fatal message and non-zero exit).
- Re-run `go test ./...` in `daemon/`.

PR #1099, opened 2026-02-17
fix: batch sweep bug-label issues across gateway and daemon

### LINKED ISSUE #588: bug(gateway): RateLimiter sessionWindows map grows unbounded
## Problem

In `gateway/src/ratelimit/index.ts`, the `sessionWindows` Map never evicts entries for sessions that stop sending requests. Over time, this causes unbounded memory growth proportional to the number of unique session keys ever seen.

The `check()` method prunes timestamps within existing entries, but never deletes entries that have zero remaining timestamps after pruning.

## File
`gateway/src/ratelimit/index.ts`

## Suggested Fix

After pruning timestamps in `check()`, delete the entry if the timestamp array is empty:

```typescript
timestamps = timestamps.filter((t) => t > cutoff);
if (timestamps.length === 0) {
  this.sessionWindows.delete(sessionKey);
  // still check global and record below
}
```

Alternatively, add a periodic sweep or LRU eviction.

## Impact
P1 — In long-running gateway deployments with many distinct sessions, memory will grow indefinitely.

### LINKED ISSUE #591: security(daemon): readJSON request body limit is 1MB but no route-level content-length check
## Problem

In `daemon/internal/api/server.go`, `readJSON` uses `io.LimitReader(r.Body, 1<<20)` (1MB limit), which is good. However, it reads the entire body into memory before parsing:

```go
raw, err := io.ReadAll(body)
```

This means every POST endpoint allocates up to 1MB per request. For endpoints like `/pairing/codes` where the expected payload is <100 bytes, this is wasteful.

More importantly, the HTTP server has no global `ReadTimeout` or `WriteTimeout` set:

```go
httpServer := &http.Server{
    Addr:    listenAddr,
    Handler: handler,
}
```

Without timeouts, a slow-loris attack can hold connections open indefinitely, exhausting file descriptors.

## File
`daemon/cmd/agentd/main.go` — `http.Server` creation

## Suggested Fix

Add standard timeouts:

```go
httpServer := &http.Server{
    Addr:         listenAddr,
    Handler:      handler,
    ReadTimeout:  15 * time.Second,
    WriteTimeout: 30 * time.Second,
    IdleTimeout:  60 * time.Second,
}
```

## Impact
P1 — Without timeouts, the daemon is vulnerable to connection exhaustion (slow-loris) attacks, especially when exposed on non-loopback interfaces.

### LINKED ISSUE #602: security(daemon): HTTP server has no read/write/idle timeouts
## Problem

`daemon/cmd/agentd/main.go` creates an `http.Server` with no timeout configuration:

```go
httpServer := &http.Server{
    Addr:    listenAddr,
    Handler: handler,
}
```

Without `ReadTimeout`, `ReadHeaderTimeout`, `WriteTimeout`, and `IdleTimeout`, a slow or malicious client can:
- Hold connections open indefinitely (slowloris attack)
- Exhaust file descriptors and goroutines
- Cause OOM by slowly feeding headers

## Suggested Fix

```go
httpServer := &http.Server{
    Addr:              listenAddr,
    Handler:           handler,
    ReadHeaderTimeout: 10 * time.Second,
    ReadTimeout:       30 * time.Second,
    WriteTimeout:      60 * time.Second,
    IdleTimeout:       120 * time.Second,
}
```

Even on loopback-only, these should be set as defense-in-depth (especially since the daemon can bind to non-loopback with an API token).

## File
- `daemon/cmd/agentd/main.go` line ~90

### LINKED ISSUE #603: bug(gateway): HttpDaemonClient has no request timeout
## Problem

`gateway/src/daemon/http_client.ts` makes HTTP requests to the daemon without any timeout:

```typescript
const response = await this.fetchImpl(url, {
    method,
    headers: { ... },
    body: body === undefined ? undefined : JSON.stringify(body),
});
```

If the daemon hangs or is unresponsive, the gateway will block indefinitely on every command. This affects all user-facing commands (`/start`, `/stop`, `/status`, etc.).

## Suggested Fix

Use `AbortSignal.timeout()` (available in Bun):

```typescript
const response = await this.fetchImpl(url, {
    method,
    headers: { ... },
    body: body === undefined ? undefined : JSON.stringify(body),
    signal: AbortSignal.timeout(30_000), // 30s timeout
});
```

Consider making the timeout configurable via `CARRIER_DAEMON_TIMEOUT_MS` env var.

## File
- `gateway/src/daemon/http_client.ts` — `requestJSON` method

### LINKED ISSUE #604: bug(ci): release workflow triggers on every main push, creating noise releases
## Problem

`release.yml` triggers on every push to `main`:

```yaml
on:
  push:
    branches:
      - main
```

Combined with the release step that always publishes:

```yaml
tag_name: main-${{ github.sha }}
```

This means **every single merge to main creates a GitHub release**. This pollutes the releases page and wastes CI minutes on building packages for non-release commits.

## Suggested Fix

Option A: Only trigger on tags:
```yaml
on:
  push:
    tags:
      - 'v*'
```

Option B: Keep the build on main push (for testing), but gate the `release` job on tag presence:
```yaml
release:
    needs: build-packages
    if: startsWith(github.ref, 'refs/tags/v')
```

## File
- `.github/workflows/release.yml`

### LINKED ISSUE #614: bug(gateway): enforce /command request body size limits + boundary tests
## Problem
`gateway/src/server.ts` parses `/command` request bodies with `request.json()` / `request.text()` but does not enforce any maximum payload size before buffering. A client can send very large bodies and force high memory use/GC pressure in the Bun process.

## Why it matters
- This is an easy denial-of-service vector on the public gateway HTTP surface.
- Resource spikes here can degrade command handling latency for normal users.
- Daemon-side code already has explicit payload-size thinking; gateway should have similar guardrails for parity.

1. Introduce a shared request-size guard for `/command` (e.g., `MAX_COMMAND_BODY_BYTES`, configurable via env with safe default).
2. Reject oversized payloads early with a stable error contract (e.g., HTTP 413 + `E_PAYLOAD_TOO_LARGE`).
3. Refactor command parsing to avoid duplicate JSON/text branches where practical and keep one validation path for "non-empty input".
4. Add focused tests in `gateway/src/server.test.ts` for:
   - payload exactly at limit (accepted)
   - payload over limit (rejected with 413 + error code)
   - both `application/json` and `text/plain` paths
   - malformed JSON under limit still returns usage error (existing behavior preserved)

## Acceptance criteria
- [ ] `/command` enforces a deterministic maximum body size.
- [ ] Oversized requests return `413` with documented JSON error shape.
- [ ] Existing valid command requests continue to work unchanged.
- [ ] New tests cover limit boundaries for JSON + text payloads.
- [ ] README/ARCHITECTURE docs mention the new limit and override mechanism (if configurable).

```bash
cd gateway
bun test src/server.test.ts
bun test
```

### LINKED ISSUE #620: bug(gateway): rate limiter is never instantiated in default server startup
## Problem

`RateLimiter` is implemented in `gateway/src/ratelimit/index.ts` and checked in `handleCommand` (`gateway/src/index.ts`), but the default `createRuntimeDependencies` in `server.ts` sets `rateLimiter` to `overrides.rateLimiter ?? undefined`:

```typescript
export function createRuntimeDependencies(overrides = {}): GatewayDependencies {
  return {
    daemon: overrides.daemon ?? new HttpDaemonClient(),
    sessions: overrides.sessions ?? new SessionStore(),
    downloads: overrides.downloads ?? new DownloadTokenStore().startPeriodicCleanup(),
    rateLimiter: overrides.rateLimiter,  // undefined by default
  };
}
```

Since `handleCommand` checks `if (deps.rateLimiter)`, rate limiting is effectively **disabled** for all production deployments unless the caller explicitly provides one.

## Location

- `gateway/src/server.ts` — `createRuntimeDependencies`
- `gateway/src/index.ts` — rate limit check in `handleCommand`

## Impact

- No rate limiting in production — any paired session can spam commands without throttling
- Combined with #617 (sessions never expire), a compromised session token has unlimited access

## Suggested fix

Instantiate a `RateLimiter` with env-based config by default:

```typescript
rateLimiter: overrides.rateLimiter ?? new RateLimiter(),
```

The `RateLimiter` constructor already reads `CARRIER_RATE_LIMIT_*` env vars with sensible defaults.

### LINKED ISSUE #633: security(daemon): Go 1.21 is end-of-life — no security patches since August 2024
## Problem

`daemon/go.mod` specifies `go 1.21`, which reached end-of-life in August 2024. The Go team no longer backports security fixes to 1.21.x.

**File:** `daemon/go.mod`

```
module carrier/daemon
go 1.21
```

This means any vulnerabilities in the Go standard library (net/http, crypto/tls, encoding/json, etc.) discovered after August 2024 are unpatched in builds using this toolchain.

## Impact

- Known CVEs in Go < 1.22 may affect the daemon (e.g., net/http request smuggling, crypto/tls issues)
- CI (`go-version-file: daemon/go.mod`) also uses this version, so test coverage runs on an outdated toolchain

## Suggested Fix

Bump to Go 1.22 or 1.23 (current supported releases):

```
go 1.23
```

Also consider adding a `toolchain` directive (Go 1.21+ feature) to pin the exact patch version.

### LINKED ISSUE #635: bug(daemon): lifecycle command execution has no per-operation timeout
## Problem

`ShellRunner.Run()` uses `exec.CommandContext(ctx)` but the callers in `lifecycle/install.go`, `start_stop.go`, and `upgrade.go` pass through the HTTP request context, which has **no deadline** (the HTTP server has no read/write timeouts — see #602).

This means install, start, stop, and upgrade commands can run **indefinitely**. A hung command blocks the goroutine and the HTTP connection forever.

**Files:**
- `daemon/internal/commandexec/runner.go` — `ShellRunner.Run` accepts context but doesn't enforce its own timeout
- `daemon/internal/lifecycle/install.go:16` — `s.runner.Run(ctx, m.Runtime.Install.Command)`
- `daemon/internal/lifecycle/start_stop.go:30` — `s.runner.Run(ctx, m.Runtime.Start.Command)`
- `daemon/internal/lifecycle/upgrade.go:63` — `s.runner.Run(ctx, m.Runtime.Upgrade.Command)`

## Impact

A malformed or hanging manifest command (e.g., a start script that blocks on stdin, a network download that never completes) will:
- Block the calling goroutine permanently
- Keep the agent in an inconsistent state (neither running nor stopped)
- Potentially exhaust goroutines under repeated calls

## Suggested Fix

Add a configurable per-operation timeout (e.g., `CARRIER_COMMAND_TIMEOUT` defaulting to 5m):

```go
func (s *Service) Install(ctx context.Context, agentID string) error {
    opCtx, cancel := context.WithTimeout(ctx, s.commandTimeout)
    defer cancel()
    result, runErr := s.runner.Run(opCtx, m.Runtime.Install.Command)
    // ...
}
```

Note: This is distinct from #602 (HTTP server timeouts) and #603 (gateway HTTP client timeout). Even with HTTP timeouts, the spawned process would continue running in the background.

Keith-CY:
## Summary
- enable default gateway rate limiting and periodic cleanup in runtime deps
- enforce `/command` request size limits (`CARRIER_MAX_COMMAND_BODY_BYTES`) with deterministic `413 E_PAYLOAD_TOO_LARGE` handling and boundary tests
- add daemon client HTTP timeout support (`CARRIER_DAEMON_TIMEOUT_MS`) using `AbortSignal.timeout`
- add daemon lifecycle per-operation command timeout (`CARRIER_COMMAND_TIMEOUT`, default `5m`) for install/upgrade with regression tests
- harden daemon JSON body parsing with early oversize rejection
- configure daemon HTTP server read/read-header/write/idle timeouts
- move release workflow to tag-driven `v*` publishing to stop noisy main-branch releases
- bump daemon toolchain target to Go 1.23 with explicit toolchain directive
- update docs to reflect new limits/timeouts and tag-driven release policy

## Issues addressed in this PR
Fixes #635
Fixes #633
Fixes #620
Fixes #614
Fixes #604
Fixes #603
Fixes #602
Fixes #591
Fixes #588

## Bug-label sweep notes
- Reviewed all open `bug` label issues in the repo.
- Items #636, #618, #617, and #590 are already functionally covered in current `main` and were not reimplemented here.
- #598 appears obsolete against the current catalog install flow (no linux/amd64 fallback checksum branch remains in `main`).

## Verification
- `cd gateway && bun test src/server.test.ts -t "runtime dependencies"`
- `cd gateway && bun test src/server.test.ts -t "command route payload size limits"`
- `cd gateway && bun test src/daemon/http_client.test.ts src/ratelimit/index.test.ts`
- `cd daemon && go test ./internal/lifecycle`
- `cd daemon && go test ./internal/api`
- `cd daemon && go test ./cmd/agentd`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,8 +2,8 @@ name: Release
 
 on:
   push:
-    branches:
-      - main
+    tags:
+      - "v*"
   workflow_dispatch:
 
 permissions:
@@ -11,14 +11,13 @@ permissions:
 
 env:
   RELEASE_ARTIFACT_RETENTION_DAYS: "14"
-  MAIN_RELEASE_KEEP_COUNT: "20"
 
 jobs:
   e2e-deploy:
     name: End-to-End Tests (Deployment)
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/'))
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     steps:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
@@ -166,7 +165,7 @@ jobs:
     timeout-minutes: 30
     permissions:
       contents: write  # Required for creating GitHub releases
-    if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/'))
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     steps:
       - name: Download packages
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093
@@ -177,46 +176,10 @@ jobs:
         uses: softprops/action-gh-release@a06a81a03ee405af7f2048a818ed3f03bbf83c7b
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          tag_name: main-${{ github.sha }}
-          name: "main-${{ github.sha }}"
+          tag_name: ${{ github.ref_name }}
+          name: ${{ github.ref_name }}
           generate_release_notes: true
           files: |
             dist/**/*.zip
             dist/**/*.sha256
             dist/**/*.sig
-
-  prune-main-releases:
-    name: Prune Old main-* Releases
-    needs: release
-    runs-on: ubuntu-latest
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-    steps:
-      - name: Prune by count policy
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          KEEP_COUNT: ${{ env.MAIN_RELEASE_KEEP_COUNT }}
-        shell: bash
-        run: |
-          set -euo pipefail
-
-          mapfile -t tags < <(
-            gh api "repos/${GITHUB_REPOSITORY}/releases" --paginate \
-              --jq '.[] | select(.tag_name | startswith("main-")) | [.created_at, .tag_name] | @tsv' \
-              | sort -r \
-              | awk -F '\t' '{print $2}'
-          )
-
-          keep="${KEEP_COUNT}"
-          total="${#tags[@]}"
-
-          echo "Found ${total} main-* releases. Keep newest ${keep}."
-
-          if (( total <= keep )); then
-            echo "No release pruning required."
-            exit 0
-          fi
-
-          for tag in "${tags[@]:keep}"; do
-            echo "Deleting release/tag: ${tag}"
-            gh release delete "${tag}" --repo "${GITHUB_REPOSITORY}" --yes --cleanup-tag
-          done
--- a/README.md
+++ b/README.md
@@ -154,9 +154,12 @@ Optional local flow from repo root:
 
 Runtime environment variables:
 - `CARRIER_DAEMON_BASE_URL` (default: `http://127.0.0.1:9090`)
+- `CARRIER_DAEMON_TIMEOUT_MS` (default: `30000`)
+- `CARRIER_COMMAND_TIMEOUT` (default: `5m`, daemon lifecycle install/upgrade command timeout)
 - `CARRIER_GATEWAY_HOST` (default: `127.0.0.1`)
 - `CARRIER_GATEWAY_PORT` (default: `8787`)
 - `CARRIER_GATEWAY_API_TOKEN` (optional on loopback; required for non-loopback bind)
+- `CARRIER_MAX_COMMAND_BODY_BYTES` (default: `65536`)
 
 #### Merge Queue
 - This repository supports GitHub Merge Queue for `main`.
--- a/daemon/cmd/agentd/agents_path_test.go
+++ b/daemon/cmd/agentd/agents_path_test.go
@@ -108,8 +108,8 @@ func TestAgentsPathHTTPRouting(t *testing.T) {
 		// Empty ID segment (double slash) → 301 (ServeMux cleans path)
 		{name: "empty_id_segment", method: "GET", path: "/api/v1/agents//status", wantStatus: http.StatusMovedPermanently},
 
-		// Special characters → 301 for traversal (ServeMux cleans ..), 404 for others
-		{name: "traversal_id", method: "GET", path: "/api/v1/agents/..%2F..%2Fetc/status", wantStatus: http.StatusMovedPermanently},
+		// Special characters → traversal payload is rejected as not found
+		{name: "traversal_id", method: "GET", path: "/api/v1/agents/..%2F..%2Fetc/status", wantStatus: http.StatusNotFound},
 		{name: "slash_encoded_id", method: "GET", path: "/api/v1/agents/a%2Fb/status", wantStatus: http.StatusNotFound},
 		{name: "leading_dot_id", method: "GET", path: "/api/v1/agents/.hidden/status", wantStatus: http.StatusNotFound},
 
--- a/daemon/cmd/agentd/main.go
+++ b/daemon/cmd/agentd/main.go
@@ -32,9 +32,13 @@ import (
 )
 
 const (
-	shutdownTimeout = 30 * time.Second
-	defaultLogsTail = 200
-	maxLogsTail     = 1000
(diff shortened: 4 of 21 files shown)


dev01lay2 (approved):
Clean, well-structured PR. Tag-driven releases, HTTP server timeouts, command timeout propagation, payload size limits, and memory cleanup for rate limiter — all solid improvements with thorough test coverage. One minor NBS inline.

Linn-San (approved):
Looks good overall — the follow-up hardening is solid.

I verified the scope changes in daemon readiness/startup safety, command timeouts, JSON size/timeout controls, and payload-size enforcement in gateway, with corresponding tests.

No blocking issues found.

PR #1107, opened 2026-02-17
fix: harden runtime decodeBody JSON parsing

Keith-CY:
## Summary
- addresses post-merge review on #1099: https://github.com/Keith-CY/carrier/pull/1099#discussion_r2816248228
- hardens runtime decodeBody used by legacy/active buildHTTPMux handlers
- enforces max body checks via both Content-Length and read cap
- rejects trailing JSON data via second decode (EOF check)
- keeps strict unknown-field rejection to preserve current API contract
- adds regression tests for trailing primitive payload, whitespace-only body, and oversized Content-Length

## Testing
- go test ./cmd/agentd
- go test ./internal/api

### DIFF
--- a/daemon/cmd/agentd/main.go
+++ b/daemon/cmd/agentd/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"crypto/subtle"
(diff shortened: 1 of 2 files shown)


dev01lay2 (approved):
Clean hardening of `decodeBody` — early Content-Length rejection, explicit empty-body check, and the `dec.Decode(&trailing)` approach for detecting trailing garbage is more robust than `dec.More()`. Tests cover all new branches well including the oversized Content-Length header case. LGTM.

daemon/cmd/agentd/main.go
```diff
@@ -491,16 +492,37 @@ type agentIDBody struct {
...
+		writeJSONError(w, http.StatusBadRequest, fmt.Sprintf("request body too large: max %d bytes", maxBodySize))
+		return false
+	}
+	if len(bytes.TrimSpace(raw)) == 0 {
+		writeJSONError(w, http.StatusBadRequest, "invalid request body: empty body")
+		return false
+	}
+
+	dec := json.NewDecoder(bytes.NewReader(raw))
 	dec.DisallowUnknownFields()
```
Akane-CN: NBS: Great hardening on `decodeBody` — combining Content-Length precheck, capped read, empty-body rejection, and explicit trailing-data detection materially improves request parsing safety.

PR #1147, opened 2026-02-17
docs: refresh carrier usage, API contract, and deployment guide

Keith-CY:
## Summary
- align README command-ingress docs with current session-token auth flow
- refresh command + daemon API contract docs to match current `/api/v1` behavior and compatibility routes
- rewrite production deployment guide for actual Go daemon + Bun gateway systemd setup

## Validation
- `bash scripts/check-doc-command-sync.sh`
- `./scripts/run-all-tests.sh` *(1 known baseline failure not caused by this PR: `gateway/src/artifact_root.test.ts` -> expected 200, received 404 in ARTIFACT_ROOT symlink path case)*

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -230,6 +230,7 @@ Restore flow:
 2. Deploy updated gateway source to `/opt/carrier/gateway`
 3. Reinstall gateway deps if lockfile changed:
    - `cd /opt/carrier/gateway && bun install --frozen-lockfile --no-progress`
+   - If this fails due to lockfile drift, regenerate and commit `gateway/bun.lock` from source before release.
 4. Restart services:
    - `sudo systemctl restart carrier-agentd carrier-gateway`
 5. Verify health and smoke-test `/pair` + `/agents`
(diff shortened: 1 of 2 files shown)


docs/deployment.md
```diff
@@ -81,127 +129,115 @@ Wants=network-online.target
...
 [Install]
 WantedBy=multi-user.target
 ```
 
-### Enable and Start
+Enable and start:
 
 ```bash
 sudo systemctl daemon-reload
 sudo systemctl enable --now carrier-agentd carrier-gateway
```
dev01lay2: NBS: `bun install --frozen-lockfile` can fail if the lockfile is stale after a source update. Might be worth a note to re-generate the lockfile during release prep.

dev01lay2 (approved):
Re-reviewed after 68bf798. The new commit is cosmetic (table alignment in error codes). Regarding Akane-CN's BS about `GET /api/v1/agents` response shape: I checked the daemon source — `ListAgents()` returns `[]AgentState` and `writeJSON` serializes it directly as a top-level array. The doc is correct as written. LGTM 👍

PR #1205, opened 2026-02-18
ci: fix release triggers and PR package comments

Keith-CY:
## Summary
- trigger release packaging on published GitHub Release events
- trigger prerelease packaging on push to main
- trigger PR test packaging for PRs targeting main and upsert package links in PR comments

## Verification
- YAML parse check:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,8 @@ on:
   workflow_dispatch:
 
 concurrency:
+  # Keep only the newest run per PR/ref/tag. Older runs can be safely cancelled
+  # because assets are keyed by the current commit and release uploads overwrite files.
   group: release-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref || github.event.release.tag_name }}
   cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,19 +1,127 @@
...
+  pull_request:
+    branches:
+      - main
+    types: [opened, reopened, synchronize]
   workflow_dispatch:
 
+concurrency:
+  group: release-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref || github.event.release.tag_name }}
+  cancel-in-progress: true
+
```
dev01lay2: NBS: Consider adding a brief comment explaining why `cancel-in-progress: true` is safe here (i.e., release assets are idempotent / overwritten anyway). Helps future readers understand the intent.

.github/workflows/release.yml
```diff
@@ -1,19 +1,127 @@
...
   push:
     branches:
       - main
+  pull_request:
+    branches:
+      - main
+    types: [opened, reopened, synchronize]
   workflow_dispatch:
 
+concurrency:
```
dev01lay2: 🟡 **Concurrency group for `release` event.** `github.event.release.tag_name` is empty for non-release events, but `github.event.pull_request.number` and `github.ref` are used as fallbacks via `||`, so this should be fine. Just confirming the precedence is intentional — PRs get `pr-number`, pushes get `ref`, releases get `tag_name`.

.github/workflows/release.yml
```diff
@@ -32,7 +140,9 @@ jobs:
         run: ./scripts/run-e2e-tests.sh
 
   build-packages:
-    needs: e2e-deploy
+    name: Build Packages (${{ matrix.label }})
```
dev01lay2: ✅ Good: `always() && needs.prepare.result == 'success'` correctly gates on prepare success while allowing e2e-deploy to be skipped for PRs.

Keith-CY:
Addressed the remaining review nits:

- Switched PR metadata generation to jq -n for robust JSON construction
- Added a brief comment explaining why cancel-in-progress: true is safe in this workflow

All checks still pass locally (check-action-pinning.sh + YAML parse).

@Akane-CN could you please take another look and re-review?

PR #1213, opened 2026-02-19
fix(ci): coverage report showing 0% — artifact download path missing

dev01lay2:
## Problem

Coverage Comment job always reports 0.00% because `download-artifact@v4` strips the common directory prefix by default. The artifact contains `daemon/coverage-func.txt`, but after download it lands as `coverage-func.txt` at workspace root. The parser looks for `daemon/coverage-func.txt` → not found → returns 0.0.

## Fix

Add `path: daemon` to the download step so files extract into `daemon/` directory where the parser expects them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,6 +203,7 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: daemon-coverage
+          path: daemon
 
       - name: Collect base branch coverage
         run: |


Akane-CN (approved):
NBS: Good targeted fix. Adding  to the coverage artifact download correctly preserves the expected file location () and resolves the false 0% report without changing broader workflow behavior. Approved.

Akane-CN (approved):
Reviewed. CI is broadly green for relevant gates, and this is a narrowly-scoped workflow fix for the coverage path regression.

I see no blocking issues:
- `actions/download-artifact@v4` now writes the artifact into `path: daemon`, which aligns with the later `daemon/coverage-func.txt` parser expectation in this same job.
- No behavior/runtime code is changed; only CI plumbing.

Approved.

Issue #1219, opened 2026-02-19
Shared memory RW access lacks authorization check


dev01lay2:
## Problem

`Policy.ResolveAccessMode` for shared memory directly honors `AccessReadWrite` requests without any authorization gate:

```go
if t == TypeShared {
    if requested == AccessReadWrite {
        return AccessReadWrite  // No auth check!
    }
    return AccessReadOnly
}
```

The code comment says "writable only with explicit authorization" but no actual authorization mechanism exists. Any agent can request RW access to any shared memory.

## Suggested Fix

Either:
1. Add an ACL/allowlist mechanism (e.g. per-memory `writers` field in manifest) and check it in `ResolveAccessMode` or `CheckMount`
2. Enforce shared memory as always RO until an authorization system is built

## Files

- `daemon/internal/memory/policy.go` — `ResolveAccessMode`

Keith-CY:
已按 **bug** 类处理，关联 PR: #1260。该 PR 合并后会自动关闭本 issue。

PR #1225, opened 2026-02-19
security(memory): enforce shared memory as read-only until ACL exists

Akane-CN:
Implements issue #1219 with fail-closed semantics for shared memory access.\n\n## Changes\n- : shared memory now always resolves to \n- Updated policy tests for shared RW requests to expect \n- Updated coverage test to assert fail-closed behavior\n\n## Why\nCurrent code path allowed shared RW without any authorization gate. This change removes that unsafe path until explicit ACL/writer authorization is implemented.\n\n## Validation\n- ok  	carrier/daemon/internal/memory	(cached)

### DIFF
--- a/daemon/internal/memory/coverage_test.go
+++ b/daemon/internal/memory/coverage_test.go
@@ -1332,15 +1332,15 @@ func TestListFilesEmpty(t *testing.T) {
 
 // ── AttachMemory with explicit mode ──
 
-func TestAttachMemorySharedExplicitRW(t *testing.T) {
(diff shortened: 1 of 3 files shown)


Linn-San (approved):
Re-reviewed #1225.

Checks are green across build/lint/tests, and the change set is small and security-oriented:
- enforces read-only default for shared memories via `ResolveAccessMode`,
- keeps per-agent semantics intact while keeping ownership/rule enforcement in mount validation,
- adds corresponding policy tests.

No blocking issues found in this pass.

Approved.

Issue #1247, opened 2026-02-20
docs: add current architecture index and retire stale TODO/TBD markers


Akane-CN:
Automated hourly architecture audit detected stale planning markers in docs without a canonical current architecture index.

Repository: Keith-CY/carrier
Default branch: main

Findings:
- docs files with Superseded/diverged markers: 4
- docs files containing TODO/TBD/FIXME: 1

Requested action:
1) Create docs/current-architecture.md as source-of-truth index.
2) Link superseded/diverged docs to canonical sections.
3) Remove/resolve stale TODO/TBD/FIXME markers or convert them into tracked issues.

This issue was created by hourly architecture audit script.

Keith-CY:
已按 **docs** 类处理，关联 PR: #1259。该 PR 合并后会自动关闭本 issue。

PR #1249, opened 2026-02-20
docs: add canonical current architecture index

Akane-CN:
Introduces  as canonical architecture index to reduce plan/implementation drift.

Refs #1247

## Acceptance mapping
- Canonical index added.
- Future superseded/diverged docs can link to a single source-of-truth.
- Provides immediate baseline for docs freshness workflow.


dev01lay2 (changes requested):
BS: This PR (#1249) is a duplicate of #1248, which is already open and approved. Same file, same content — only the generated timestamp differs. Please close this one to avoid merge conflicts and churn.

PR #1260, opened 2026-02-21
fix(memory): harden attachment validation and prepare concurrency

### LINKED ISSUE #1218: Race condition in PrepareAgentMemory — concurrent prepare for same agent unsafe
## Problem

In `view.go`, `PrepareAgentMemory` releases `RLock` mid-flow then calls `MountsForAgent` (which re-acquires `RLock`), and later calls `applyMountStateWithRollback` (which acquires write lock). Between these lock acquisitions, another goroutine could prepare the same agent concurrently, leading to:

- Corrupted effective view directory (two goroutines writing to same `views/<agentID>/effective/`)
- Inconsistent mount state (one goroutine rolls back mounts the other just applied)

## Suggested Fix

Add per-agent granularity locking (e.g. `sync.Map` of `*sync.Mutex` keyed by agentID) so that `PrepareAgentMemory` serializes per agent while allowing concurrent prepares for different agents.

## Files

- `daemon/internal/memory/view.go` — `PrepareAgentMemory`

### LINKED ISSUE #1219: Shared memory RW access lacks authorization check

### LINKED ISSUE #1242: memory: reject archived entries in SetAttachmentsFromLinks
## Summary
`SetAttachmentsFromLinks` currently validates not-found, owner mismatch, and per-agent limits, but it does **not** reject archived memories. `AttachMemory` already blocks archived entries, so link-based attachment can bypass that lifecycle guard.

## Why this matters
- Inconsistent lifecycle policy between attachment paths.
- Archived memory can be reintroduced through link sync flows.
- Confuses operators and weakens state model expectations.

## Scope
- Add archived-state validation to `SetAttachmentsFromLinks` consistent with `AttachMemory`.
- Add unit coverage for archived rejection in link-based attach flow.
- Update runtime/memory docs to reflect this enforced guard.

## Acceptance
- [ ] `SetAttachmentsFromLinks` returns an error when any target memory is archived.
- [ ] Existing not-found/owner/per-agent checks continue to work.
- [ ] Tests cover archived rejection path.
- [ ] Docs updated in the same PR.

Keith-CY:
## Summary
- reject archived memories in `SetAttachmentsFromLinks` to align with `AttachMemory` lifecycle guard
- serialize `PrepareAgentMemory` per agent to avoid concurrent compose/rollback races
- keep shared memory fail-closed as read-only and remove lingering core TODO marker
- add coverage for archived-link rejection and concurrent prepare behavior

## Validation
- go test ./internal/memory
- go test ./...

Closes #1242
Closes #1218
Closes #1219

### DIFF
--- a/daemon/internal/memory/mempack_test.go
+++ b/daemon/internal/memory/mempack_test.go
@@ -343,6 +343,66 @@ func TestStorePersistsStateAcrossRestart(t *testing.T) {
 	}
 }
 
+func TestPrepareAgentMemoryConcurrentSameAgent(t *testing.T) {
+	store, _ := newMemoryStoreWithRoot(t)
+	pack := filepath.Join(t.TempDir(), "shared.mempack.zip")
+	writeMempack(t, pack, baseManifest("shared"), map[string]string{
+		"content/prompts/system.md": "shared",
+		"content/kb/faq.txt":        "faq",
+	})
+
+	entry, err := store.ImportMemory(pack, ImportOptions{TargetRegion: TypeShared})
+	if err != nil {
+		t.Fatalf("import shared: %v", err)
+	}
+	if _, err := store.AttachMemory("agent-1", entry.ID, AttachOptions{Priority: 10}); err != nil {
+		t.Fatalf("attach shared: %v", err)
+	}
+
+	const workers = 8
+	start := make(chan struct{})
+	var wg sync.WaitGroup
+	errs := make(chan error, workers)
+	digests := make(chan string, workers)
+
+	for i := 0; i < workers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-start
+			contract, err := store.PrepareAgentMemory("agent-1")
+			if err != nil {
+				errs <- err
+				return
+			}
+			digests <- contract.ViewDigest
+		}()
+	}
+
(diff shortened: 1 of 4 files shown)


Linn-San (approved):
Re-reviewed #1260.

Checks pass. This is a continuation/refinement of the memory read-only/concurrency work with improved validation and concurrent prepare coverage:
- rejects archived memory in attachment resolution,
- adds deterministic per-agent prepare serialization,
- adds concurrency and regression tests to lock down behavior.

I see no blocking regressions.

Approved.

PR #1261, opened 2026-02-21
feat(memory): persistence, incremental prepare reuse, and export guards

### LINKED ISSUE #1217: Memory store lacks persistence — data lost on daemon restart
## Problem

The entire `memory.Store` is a pure in-memory map (`entries`, `mounts`, `attachments`, `manifests`, `installPath`). When the daemon restarts, all memory state is lost even though the package files remain on disk.

## Impact

- Agents lose their memory attachments after any daemon restart
- Mount state and audit history are ephemeral
- In desktop environments where restarts are frequent (app updates, sleep/wake), this is especially problematic

## Suggested Approaches

1. **Scan-and-rebuild on startup**: Walk `packages/` directory tree, re-parse each `memory.yaml` manifest, reconstruct `entries` + `manifests` + `installPath` maps. Attachments would need a separate persistence file (e.g. `attachments.json` per agent or a global state file).
2. **SQLite backing store**: Replace in-memory maps with SQLite tables. Gives ACID, crash safety, and queryability.
3. **WAL from audit events**: Since `recordAudit` already captures all mutations, persist audit events to a file and replay on startup to reconstruct state.

## Files

- `daemon/internal/memory/store.go` — Store struct with in-memory maps
- `daemon/internal/memory/view.go` — PrepareAgentMemory, attachments
- `daemon/internal/lifecycle/memory.go` — autoMount/autoUnmount integration

### LINKED ISSUE #1220: PrepareAgentMemory does full file copy every time — performance concern for large packages
## Problem

`PrepareAgentMemory` in `view.go` calls `os.RemoveAll(effectiveDir)` then copies every file from every attached memory package on each invocation. For large memory packages this will be slow and I/O intensive, especially on desktop where SSDs have write endurance limits.

## Suggested Improvements

1. **Use hardlinks or symlinks** for read-only mounts instead of copying files
2. **Incremental updates**: Compare previous view digest with current; only re-compose if attachments changed
3. **Copy-on-write**: Only copy when an agent needs to modify a file (for RW mounts)

## Files

- `daemon/internal/memory/view.go` — `PrepareAgentMemory`, `copyFile`

### LINKED ISSUE #1221: ExportMemory has no size limit or concurrency control
## Problem

`ExportMemory` in `mempack_store.go` zips an entire memory package to `artifacts/exports/` without:
- Maximum artifact size check
- Concurrent export limiting
- Disk space pre-check

A large package or multiple concurrent exports could exhaust disk space or cause OOM.

## Suggested Fix

1. Add a configurable max export size (reject if package exceeds limit)
2. Use a semaphore to limit concurrent exports (e.g. `make(chan struct{}, 3)`)
3. Check available disk space before starting export
4. Consider streaming the zip to the HTTP response instead of writing to disk

## Files

- `daemon/internal/memory/mempack_store.go` — `ExportMemory`

### LINKED ISSUE #1223: Desktop platform compatibility: storage paths, permissions, and path separators
## Problem

The memory system assumes a Linux server environment. For desktop deployment (macOS/Windows/Linux desktop), several issues arise:

### 1. Hardcoded storage paths
- `runtimeReadTargetPath` defaults to `/app/memory`
- `runtimeWriteTargetPath` defaults to `/app/memory_private`
- `rootDir` is expected to be set but no platform-aware default exists

**Fix**: Use platform-appropriate defaults:
- macOS: `~/Library/Application Support/carrier/memory/`
- Linux: `$XDG_DATA_HOME/carrier/memory/` (default `~/.local/share/`)
- Windows: `%APPDATA%\carrier\memory\`

### 2. File permissions
- Code uses `0o755` and `0o644` throughout
- Windows ignores Unix permission bits; macOS may require different handling for sandboxed apps

**Fix**: Abstract permission handling, use platform-appropriate defaults

### 3. Memory ID uses `/` as separator
- `memoryIDFor` produces IDs like `public/publisher/id@version`
- These are used as map keys (fine) but also appear in file paths via `installPathFor`
- On Windows, ensure ID-to-path conversion is correct (it currently uses `filepath.Join` which should handle this, but the ID itself containing `/` could confuse path-based lookups)

**Fix**: Verify ID handling on Windows; consider using a delimiter that is not a path separator (e.g. `:` or `::`)

## Files

- `daemon/internal/memory/store.go` — default paths
- `daemon/internal/memory/mempack_store.go` — `memoryIDFor`, `installPathFor`, permissions
- `daemon/internal/memory/view.go` — `copyFile`, directory creation permissions

Keith-CY:
## Summary
- persist memory store state (entries/mounts/manifests/install paths/attachments/views) to disk and reload on startup
- avoid full memory-view recomposition when prepare input is unchanged by reusing existing composed view\n- add export safety guards: max export bytes, concurrency limiter, and disk-availability precheck
- move default runtime mount targets to platform-aware user config directories
- add comprehensive tests for persistence, reuse path, and export guard behavior

## Validation
- go test ./internal/memory
- go test ./...
 
- Closes #1217
- Closes #1220 
- Closes #1221 
- Closes #1223

### DIFF
--- a/daemon/internal/memory/disk_space_windows.go
+++ b/daemon/internal/memory/disk_space_windows.go
@@ -1,0 +1,9 @@
+//go:build windows
+
+package memory
+
+import "errors"
+
+func availableDiskBytes(path string) (uint64, error) {
+	return 0, errors.New("disk free-space probe unsupported on windows build")
+}
--- a/daemon/internal/memory/persistence.go
+++ b/daemon/internal/memory/persistence.go
@@ -1,0 +1,135 @@
+package memory
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type persistedStoreState struct {
+	Entries         map[string]Entry           `json:"entries"`
+	Mounts          []MountRecord              `json:"mounts"`
+	Manifests       map[string]PackageManifest `json:"manifests"`
+	InstallPath     map[string]string          `json:"install_path"`
+	Attachments     map[string][]Attachment    `json:"attachments"`
+	Views           map[string]ViewExplanation `json:"views"`
+	ViewInputDigest map[string]string          `json:"view_input_digest"`
+}
+
+func (s *Store) loadState() error {
+	statePath := strings.TrimSpace(s.statePath)
+	if statePath == "" {
+		return nil
+	}
+
+	raw, err := os.ReadFile(statePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("read memory state: %w", err)
+	}
+
+	var state persistedStoreState
+	if err := json.Unmarshal(raw, &state); err != nil {
+		return fmt.Errorf("parse memory state: %w", err)
+	}
+
+	if state.Entries != nil {
+		s.entries = state.Entries
+	}
+	if state.Mounts != nil {
+		s.mounts = state.Mounts
+	}
+	if state.Manifests != nil {
+		s.manifests = state.Manifests
+	}
+	if state.InstallPath != nil {
+		s.installPath = state.InstallPath
+	}
+	if state.Attachments != nil {
+		s.attachments = state.Attachments
+	}
+	if state.Views != nil {
+		s.views = state.Views
+	}
+	if state.ViewInputDigest != nil {
+		s.viewInputDigest = state.ViewInputDigest
+	}
+
+	return nil
+}
+
+func (s *Store) persistStateLocked() error {
+	statePath := strings.TrimSpace(s.statePath)
+	if statePath == "" {
+		return nil
+	}
+
+	state := persistedStoreState{
+		Entries:         make(map[string]Entry, len(s.entries)),
+		Mounts:          append([]MountRecord(nil), s.mounts...),
+		Manifests:       make(map[string]PackageManifest, len(s.manifests)),
+		InstallPath:     make(map[string]string, len(s.installPath)),
+		Attachments:     make(map[string][]Attachment, len(s.attachments)),
+		Views:           make(map[string]ViewExplanation, len(s.views)),
+		ViewInputDigest: make(map[string]string, len(s.viewInputDigest)),
+	}
+	for k, v := range s.entries {
+		state.Entries[k] = v
+	}
+	for k, v := range s.manifests {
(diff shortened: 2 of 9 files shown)


daemon/internal/memory/persistence.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if err := os.Rename(tmpPath, statePath); err != nil {
+		_ = os.Remove(tmpPath)
+		s.lastStateErr = fmt.Errorf("replace memory state: %w", err)
+		return s.lastStateErr
+	}
+
+	s.lastStateErr = nil
+	return nil
+}
```
dev01lay2: NBS: Consider `fsync` on the temp file before rename for crash-safety on ext4 with `data=writeback`. Current approach is fine for most production setups.

daemon/internal/memory/disk_space_windows.go
```diff
@@ -0,0 +1,9 @@
+//go:build windows
+
+package memory
+
+import "errors"
+
+func availableDiskBytes(path string) (uint64, error) {
+	return 0, errors.New("disk free-space probe unsupported on windows build")
```
dev01lay2: NBS: Could use `GetDiskFreeSpaceExW` via syscall for a real Windows implementation in a follow-up. The best-effort fallback is sensible for now.

daemon/internal/memory/view.go
```diff
@@ -112,20 +122,26 @@ func (s *Store) ListAttachments(agentID string) []Attachment {
 // The incoming order is preserved as ascending priority.
 func (s *Store) SetAttachmentsFromLinks(agentID string, memoryIDs []string) error {
 	s.mu.Lock()
-	defer s.mu.Unlock()
 
 	attachments := make([]Attachment, 0, len(memoryIDs))
 	hasPerAgent := false
 	for i, memoryID := range memoryIDs {
 		entry, ok := s.entries[memoryID]
 		if !ok {
```
dev01lay2: NBS: The manual `s.mu.Unlock()` on each early-return path (replacing the previous `defer`) is correct but increases maintenance risk. If this file grows, consider a helper that returns under lock and persists+unlocks in one place.

dev01lay2 (approved):
Thorough feature set — persistence, prepare-reuse, and export guards all look correct. Atomic write pattern for state, input-digest caching, and semaphore-based concurrency control are solid. CI green.

PR #1274, opened 2026-02-21
fix(release): preserve .zip extension for PR test package downloads

Keith-CY:
## Summary
- use the built package filename (with .zip) as the uploaded artifact name in release PR runs
- move PR test-package comment generation into scripts/ci/pr-package-comment.cjs
- add regression tests for comment generation and .zip package validation

## Why
Artifact links in PR comments point to GitHub Actions artifact downloads. When artifact names omit .zip, downloaded files can also lose the .zip suffix, causing extraction failures.

## Testing
- node --test scripts/ci/*.test.cjs

### DIFF
--- a/scripts/ci/pr-package-comment.cjs
+++ b/scripts/ci/pr-package-comment.cjs
@@ -1,0 +1,128 @@
+"use strict";
+
+const fs = require("node:fs");
+const path = require("node:path");
+
+const COMMENT_MARKER = "<!-- carrier-pr-test-packages -->";
+
+function listMetadataFiles(metadataDir) {
+  if (!fs.existsSync(metadataDir)) {
+    return [];
+  }
+
+  return fs
+    .readdirSync(metadataDir, { withFileTypes: true })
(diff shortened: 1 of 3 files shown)


scripts/ci/pr-package-comment.cjs
```diff
@@ -0,0 +1,128 @@
...
+    throw new Error(`Expected "package_file" to end with .zip in ${filePath}, got "${packageFile}"`);
+  }
+  if (!artifactId || !/^\d+$/.test(artifactId)) {
+    throw new Error(`Invalid "artifact_id" in ${filePath}: "${artifactId}"`);
+  }
+
+  return {
+    label,
+    packageFile,
+    artifactId,
```
dev01lay2: NBS: The `.zip` check is a nice safeguard. If other archive formats are ever needed (e.g. `.tar.gz`), this will need updating — but hard-coding `.zip` is the right call for now given the current build matrix.

PR #1304, opened 2026-02-21
ci(release): ship binary-only artifacts named carrier

Keith-CY:
## Summary
- update release packaging to output binary-only artifacts
- rename packaged executable from `agentd` to `carrier` (and `carrier.exe` on Windows)
- stop copying repository source directories into release archives

## Why
Current CI build artifacts are valid builds, but their package layout does not match the intended distribution shape. This change aligns release artifacts with the desired format: executable-only payload.

## Validation
- local smoke build/package (linux-x64) using workflow-equivalent commands
- verified zip content contains exactly one file: `carrier`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -199,19 +199,17 @@ jobs:
         run: |
           set -euo pipefail
           archive_name="${{ needs.prepare.outputs.package_prefix }}-${{ matrix.label }}"
-          binary_name="agentd${{ matrix.ext }}"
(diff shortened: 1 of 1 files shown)


Akane-CN (approved):
Reviewed. Build/release workflow updates look coherent and CI is green.

NBS: This change renames packaged release binary from agentd to carrier. If intentional, please update user-facing quickstart/install docs (and any downstream automation assumptions) in the same stream to avoid command mismatches after unzip.

PR #1306, opened 2026-02-21
ci(release): fix cosign keyless spot-check verification

Keith-CY:
## Summary
- update scripts/sign-artifacts.sh to emit Sigstore bundles (`*.sigstore.json`) alongside detached signatures
- update scripts/verify-signature.sh to prefer bundle-based keyless verification for cosign v2.4+, with cert-based fallback and clearer errors
- pass explicit cert identity/issuer in release spot-check step and publish/upload bundle files with release artifacts
- extend script tests for bundle behavior and stabilize existing verify-script tests in sandboxed environments

## Why
Release push runs fail in sign-artifacts -> Verify signatures (spot-check) with:

> provide a key with --key or --sk, a certificate to verify against with --certificate, or a bundle with --bundle

This change aligns signing and verification with modern keyless blob verification requirements.

## Validation
- bash scripts/test-sign-artifacts.sh
- bash scripts/test-verify-signature.sh
- bash -n scripts/sign-artifacts.sh scripts/verify-signature.sh scripts/test-sign-artifacts.sh scripts/test-verify-signature.sh

### DIFF
--- a/scripts/test-verify-signature.sh
+++ b/scripts/test-verify-signature.sh
@@ -245,6 +245,92 @@ else
     log_fail "Bundle-based cosign verification should succeed"
 fi
 
+# Test 10: Explicit signature argument takes precedence over bundle
+run_test "Test 10: Explicit signature-file should bypass bundle mode"
+mkdir -p "$TEST_DIR/mock-bin-explicit"
+cat > "$TEST_DIR/mock-bin-explicit/cosign" <<'MOCK'
+#!/usr/bin/env bash
+set -euo pipefail
+
+saw_bundle=false
+sig=""
+cert=""
+identity=""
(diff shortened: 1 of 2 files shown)


Akane-CN (changes requested):
I found a behavior mismatch that should be fixed before merge.

BS: In scripts/verify-signature.sh, when a `*.sigstore.json` bundle exists, the cosign path always prefers bundle verification, which effectively ignores an explicitly provided second `signature-file` argument. This breaks the script's documented CLI contract and can mislead callers who intend to verify a specific signature file.

Please ensure explicit `signature-file` input is honored (or clearly reject it when bundle mode is active with an explicit error), and keep behavior/documentation aligned.

PR #1333, opened 2026-02-22
fix: tolerate daemon EOF during managed-agent install/start

Keith-CY:
## Summary
- reconcile `carrier add` daemon `EOF` errors by checking daemon status for `install`/`start`
- treat EOF as success only when daemon status already reached the target state
- add tests covering EOF reconciliation success and mismatch failure

## Validation
- go test ./cmd/carrier

NBS: add an integration test that simulates stale keep-alive POST EOF against the daemon route to guard future HTTP transport changes.

### DIFF
--- a/cmd/carrier/main.go
+++ b/cmd/carrier/main.go
@@ -1582,12 +1582,112 @@ func daemonAgentAction(agentID, action string) error {
 	path := fmt.Sprintf("/api/v1/agents/%s/%s", neturl.PathEscape(agentID), neturl.PathEscape(action))
 	_, status, err := daemonRequest(http.MethodPost, path, map[string]string{})
 	if err != nil {
+		reconciled, reconcileErr := reconcileDaemonActionOnEOF(agentID, action, err)
+		if reconciled {
+			return nil
+		}
+		if reconcileErr != nil {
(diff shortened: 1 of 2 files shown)


cmd/carrier/main.go
```diff
@@ -1590,6 +1597,99 @@ func daemonAgentAction(agentID, action string) error {
...
+		} `json:"statuses"`
+	}
+	if err := json.Unmarshal(raw, &wrapped); err != nil {
+		return daemonAgentStatusSnapshot{}, fmt.Errorf("decode daemon status response: %w", err)
+	}
+	if len(wrapped.Statuses) == 0 {
+		return daemonAgentStatusSnapshot{}, errors.New("daemon status response did not include statuses")
+	}
+	first := wrapped.Statuses[0]
+	return daemonAgentStatusSnapshot{
```
dev01lay2: NBS: `isDaemonEOFError` falls back to string matching (`strings.Contains(..., "eof")`). This catches edge cases but could also match unrelated errors with 'eof' in their message. Consider logging when string-match path is hit so unexpected matches are visible in debug logs.

Akane-CN (approved):
Approved.

EOF reconciliation based on daemon status for install/start is a solid fix, and the added success/failure tests for status-matched vs mismatched EOF paths make the behavior explicit.

PR #1335, opened 2026-02-22
fix: avoid carrier add timeout on long openclaw installs

Keith-CY:
## Summary
- route managed-agent daemon POSTs through a timeout-aware helper
- extend `install` action timeout from 5 minutes to 20 minutes to handle long `openclaw` install scripts
- keep default daemon request timeout at 5 minutes for non-install actions
- add unit test coverage for timeout selection logic

## Validation
- go test ./cmd/carrier

NBS: add an end-to-end test that exercises a slow install path (>5m) and verifies `carrier add` remains stable.


Akane-CN (approved):
Approved.

Timeout-aware routing with a longer install timeout (while preserving tighter defaults for non-install actions) is the right tradeoff, and the timeout selection tests make the behavior clear.

PR #1343, opened 2026-02-22
fix(managed-add): surface daemon failure details and extend install timeout

Keith-CY:
## Summary
- surface daemon `lastError` when install/start request hits EOF so TUI no longer only reports opaque EOF
- treat broken/error install/runtime states as terminal reconciliation failures with actionable messages
- raise daemon lifecycle command timeout default from 5m to 20m for long EC2 OpenClaw builds
- add regression tests for EOF reconciliation and timeout env/default behavior

## Verification
- `go test ./cmd/carrier`
- `(cd daemon && go test ./internal/lifecycle)`

## Context
EC2 TUI validation showed `carrier add openclaw` on low-resource hosts can exceed 5m during git+pnpm build; daemon returned `signal: killed` while CLI surfaced only EOF. This patch keeps the actionable daemon error and avoids premature timeout.

### DIFF
--- a/cmd/carrier/main.go
+++ b/cmd/carrier/main.go
@@ -1587,7 +1587,7 @@ func daemonAgentAction(agentID, action string) error {
 			return nil
 		}
 		if reconcileErr != nil {
-			return fmt.Errorf("%w (status reconciliation failed: %v)", err, reconcileErr)
+			return fmt.Errorf("daemon %s %s failed: %v (request error: %w)", action, agentID, reconcileErr, err)
 		}
 		return err
 	}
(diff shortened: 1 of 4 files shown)


dev01lay2 (approved):
Good improvement over #1342 — surfacing `lastError` from daemon status gives actionable failure messages instead of opaque EOF errors. Matching the daemon-side `defaultCommandTimeout` to 20min aligns both sides. Test coverage for broken/error/crashed states is thorough. LGTM.

Linn-San (approved):
Re-reviewed #1343.

Checks are green and no runtime/test blockers are visible in the changed scope.

I reviewed the two functional adjustments:
- surfaces daemon reconciliation details on EOF by reading `lastError` from agent status for install/start
- raises default command timeout from 5m to 20m for lifecycle operations.

The new error-context path is a good reliability improvement (more actionable failure text), and I do not see incorrect timeout/default semantics.

Approved.

PR #1347, opened 2026-02-22
fix(catalog): fallback OpenClaw install on low-memory EC2

Keith-CY:
## Summary

- keep managed OpenClaw installs git-first by default
- add automatic fallback to upstream default install mode when git path fails on low-memory hosts
- allow operator override via CARRIER_OPENCLAW_INSTALL_METHOD and strict mode via CARRIER_OPENCLAW_DISABLE_INSTALL_FALLBACK
- harden install-method interpolation with token sanitization

## Testing
- go test ./internal/catalog -count=1

### DIFF
--- a/daemon/internal/catalog/manifests.go
+++ b/daemon/internal/catalog/manifests.go
@@ -38,6 +38,9 @@ const (
 	defaultDaemonHealthzPath = "/healthz"
 	defaultDaemonHealthURL   = "http://localhost:9090/healthz"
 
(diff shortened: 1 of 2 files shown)


Akane-CN (approved):
Approved.

Git-first with controlled fallback (plus strict/override env controls) is a good resilience balance for low-memory hosts, and the install-method token sanitization + catalog tests make the fallback path safer and explicit.

PR #1351, opened 2026-02-22
fix(catalog): harden openclaw installer fallback paths from gemini feedback

Akane-CN:
## Summary
- replace unix `curl | bash` execution with temp-file flow to avoid masked download failures
- run Windows fallback installer in child PowerShell process so script-level `exit` does not bypass fallback
- apply `CARGO_BUILD_JOBS=1` on explicit git install path too
- update catalog tests for new command contract and add explicit git memory-guard test

## Issues
- Closes #1348
- Closes #1349
- Closes #1350

## Verification
- cd daemon && go test ./internal/catalog -count=1

PR #1352, opened 2026-02-22
fix(catalog): make OpenClaw fallback path non-interactive

Keith-CY:
## Summary\n- change OpenClaw fallback path from implicit installer default to explicit npm mode\n- keep git-first behavior for managed installs, but avoid interactive prompt hangs when fallback triggers\n- preserve no-onboard behavior in both primary and fallback paths\n\n## Testing\n- go test ./internal/catalog -count=1

### DIFF
--- a/daemon/internal/catalog/manifests.go
+++ b/daemon/internal/catalog/manifests.go
@@ -69,15 +69,15 @@ func getInstallCommand() string {
 	switch runtime.GOOS {
 	case "windows":
 		if allowFallback {
-			// Keep git as first choice, then retry with installer default mode.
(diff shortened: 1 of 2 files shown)


dev01lay2 (changes requested):
BS: This PR conflicts with main and appears to be a less-hardened variant of the already-merged #1347 and open #1351. Key differences:

1. Fallback still pipes `curl | bash` instead of downloading to tmpfile first (#1351 fixed this)
2. Fallback uses `--install-method npm` instead of omitting install-method to let the installer pick its default
3. Missing `CARGO_BUILD_JOBS=1` on explicit git paths
4. Windows fallback doesn't use child PowerShell processes (#1351 fixed this)

Please close this in favor of #1351 which addresses Gemini's feedback on #1347 with proper hardening.

PR #1355, opened 2026-02-22
fix: add debug log when isDaemonEOFError string-match fallback is hit

### LINKED ISSUE #1334: [review-followup] PR #1333: `isDaemonEOFError` falls back to string matching (`strings.Contains(..., "eof...
Source PR: https://github.com/Keith-CY/carrier/pull/1333
PR title: fix: tolerate daemon EOF during managed-agent install/start

## Suggestion
- [ ] `isDaemonEOFError` falls back to string matching (`strings.Contains(..., "eof")`). This catches edge cases but could also match unrelated errors with 'eof' in their message. Consider logging when string-match path is hit so unexpected matches are visible in debug logs.

## Trace
- review-comment:dev01lay2

dev01lay2:
Adds a debug log to stderr when `isDaemonEOFError` falls back to string matching instead of `errors.Is(err, io.EOF)`. This makes unexpected matches visible in debug output.

Closes #1334

### DIFF
--- a/cmd/carrier/main.go
+++ b/cmd/carrier/main.go
@@ -1711,7 +1711,7 @@ func isDaemonEOFError(err error) bool {
 	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
 		return true
 	}
-	if strings.Contains(strings.ToLower(err.Error()), "eof") {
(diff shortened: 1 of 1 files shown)


Akane-CN (approved):
Approved.\n\nDebug log on string-match fallback in isDaemonEOFError improves observability with minimal risk.

PR #1358, opened 2026-02-22
fix(timeout): reconcile managed add install on daemon transport timeout

Keith-CY:
## Summary
- increase managed add install action timeout to track daemon command timeout plus safety buffer
- include transport timeout errors in daemon action reconciliation (in addition to EOF)
- add tests for timeout reconciliation and timeout window behavior

## Verification
- go test ./cmd/carrier


Akane-CN (approved):
Approved.

Good reliability follow-up: aligning managed-add install timeout with daemon timeout (+buffer) and extending reconciliation to transport timeout errors closes a real long-install failure gap. Test coverage for timeout window and reconciliation paths looks appropriate.

PR #1361, opened 2026-02-22
fix: stabilize EC2 openclaw install/start path

Keith-CY:
## Summary
- detach daemon install/upgrade operations from request cancellation so client disconnects/timeouts do not kill in-flight installs
- switch OpenClaw runtime start command from `gateway start` to `gateway` to match current OpenClaw CLI behavior
- keep catalog manifest and tests aligned with the new startup command

## Validation
- go test ./internal/api ./internal/catalog (in daemon module)
- go test ./cmd/carrier

## Why
On EC2 free-tier verification, timed-out clients could terminate install execution and leave OpenClaw in a broken state; additionally, the previous start subcommand exited immediately on current OpenClaw builds.

### DIFF
--- a/catalog/openclaw.manifest.json
+++ b/catalog/openclaw.manifest.json
@@ -17,7 +17,7 @@
       "command": "npm install -g openclaw@1.0.0"
     },
     "start": {
-      "command": "openclaw gateway start"
+      "command": "openclaw gateway"
     },
     "stop": {
       "command": "openclaw gateway stop"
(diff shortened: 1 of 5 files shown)


Akane-CN (approved):
Approved.

Decoupling install/upgrade execution from request cancellation addresses the EC2 timeout/disconnect failure mode, and switching OpenClaw start invocation to  aligns runtime behavior with current CLI semantics. Catalog/tests/docs updates are consistent with the new contract.

PR #1328, opened 2026-02-22
docs: refresh current architecture audit snapshot

Akane-CN:
Refreshes canonical architecture snapshot based on latest hourly audit.

## Acceptance mapping
- `docs/current-architecture.md` updated with latest metrics.
- Open improvement tracks linked (when any).
- Keeps docs freshness aligned with current implementation state.

PR #1367, opened 2026-02-23
docs: refresh current architecture audit snapshot

Akane-CN:
Refreshes canonical architecture snapshot based on latest hourly audit.

## Acceptance mapping
- `docs/current-architecture.md` updated with latest metrics.
- Open improvement tracks linked (when any).
- Keeps docs freshness aligned with current implementation state.


dev01lay2 (changes requested):
BS: Same recurring issue as #1273, #1328, and #1365 — this overwrites `docs/current-architecture.md` with generated audit output, destroying the curated canonical index.

The audit automation must target `docs/audit-snapshot.md` instead (as established in #1318 and fixed in the previous iteration of #1365). The underlying automation script still hasn't been permanently fixed to use the correct target file.

PR #1467, opened 2026-02-27
feat: make carrier remote add the deterministic remote install path

Keith-CY:
## Summary
- refactor CLI remote workflow to `carrier remote add` as the single deterministic install entrypoint (no fallback compatibility path)
- make gateway remote install agent-aware (`openclaw`, `picoclaw`, `zeroclaw`) and route stream/non-stream install through shared dispatcher
- support instance config get/patch API path for remote instances and wire CLI sync into instance-level config patch
- remove legacy `scripts/remote-openclaw-install.sh`; update VPS suite script and docs to use `carrier remote add`
- update chat `/install` handling to use generic `<agent_id> <host_id>` host-binding flow
- fix onboard minimal flow regressions: channel selection (including WebUI-only), credential/token reuse prompts, and no pair-code fetch in WebUI-only mode

## Verification
- `GOCACHE=/tmp/go-build-carrier go test ./cmd/carrier`
- `GOCACHE=/tmp/go-build-carrier go test ./...` (repo root module)
- `GOCACHE=/tmp/go-build-carrier go test ./...` (in `gateway/`)
- `GOCACHE=/tmp/go-build-carrier go build ./cmd/carrier`
- `GOCACHE=/tmp/go-build-carrier go build ./...` (in `gateway/`)
- `bash -n scripts/remote-vps-agent-suite.sh`

### DIFF
--- a/cmd/carrier/main.go
+++ b/cmd/carrier/main.go
@@ -2031,15 +2031,21 @@ func buildJSONRemoteConfigPatch(opts remoteCommandOptions, cfg *configv2.Config)
 }
 
 func buildZeroClawRemoteConfigPatch(opts remoteCommandOptions, cfg *configv2.Config) (map[string]interface{}, error) {
+	syncChannels := dedupeLowerStrings(opts.SyncChannels)
+	syncProviders := dedupeLowerStrings(opts.SyncProviders)
+	if (len(syncChannels) == 0) != (len(syncProviders) == 0) {
+		return nil, errors.New("zeroclaw sync requires both --sync-channel and --sync-provider to avoid overwriting existing remote config")
+	}
+
 	var (
 		defaultProvider = "openai"
 		defaultModel    = "openai/gpt-5.3-codex"
 		defaultAPIKey   = ""
 	)
-	if len(opts.SyncProviders) > 0 {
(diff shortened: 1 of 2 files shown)


dev01lay2 (approved):
Thorough refactor — `carrier remote add` as the single deterministic entrypoint is a big DX improvement over the old multi-action shell script. Multi-agent install, config sync, and the `reset` command all look solid. Docs are well-aligned with the new surface. A few non-blocking suggestions inline.

cmd/carrier/main.go
```diff
@@ -1443,121 +1477,810 @@ func runListInstances(out io.Writer) error {
...
+			"[channels_config.telegram]",
+			"bot_token = \"\"",
+			"allowed_users = []",
+			"mention_only = false",
+		)
+	}
+
+	lines := []string{
+		"# Generated by Carrier CLI remote add",
+		fmt.Sprintf("api_key = %s", strconv.Quote(defaultAPIKey)),
```
Akane-CN: BS: Same full-overwrite issue on provider fields: when `--sync-provider` is omitted, this writes `api_key = ""` and default provider/model values into the generated TOML. A channel-only sync will clobber a previously valid provider config and break runtime auth/model selection. For zeroclaw, missing provider selection should not rewrite provider credentials/model, or the command should hard-fail unless both channel+provider are explicitly provided.

PR #1495, opened 2026-02-28
feat(memory): finalize FusionMem with SQLite index + migration APIs

Keith-CY:
## Summary
- add SQLite-backed FusionMem index/cache with FTS5 search fallback
- sync records/observations/grants/instance scopes into SQLite while keeping disk truth as source of truth
- add migration lifecycle APIs: backup, validate, rollback
- extend memory v2 server endpoints with migration routes and add end-to-end tests
- add root module transitive deps required by cmd build for SQLite backend

## Validation
- `go test ./cmd/carrier -run TestNonExistent`
- `go test ./...` (daemon)
- `go test ./...` (baseagent)


dev01lay2 (approved):
SQLite index/cache layer with FTS5 search and migration lifecycle APIs looks well-structured. Disk truth as source with SQLite as cache is a solid pattern. Note: PR description shows test failures in cmd/carrier and daemon — make sure those pass before merging the stack. LGTM on the code.

PR #1497, opened 2026-02-28
security: harden Lima isolation — custom template + VM cleanup on uninstall

### LINKED ISSUE #1493: security: harden Lima isolation — restrict default mount to agent workspace only
## Problem

When Carrier uses Lima for macOS isolation (`limactl create -y`), it relies on Lima's **default template** which mounts `~` (read-only) into the VM. This means the isolated agent can read:

- `~/.ssh/` — SSH private keys
- `~/.aws/` — AWS credentials  
- `~/.config/` — various app configs and tokens
- `~/.gnupg/` — GPG keys

While the mount is read-only (Lima >= 0.17 default), **read access to secrets is already a security breach** for an isolation feature.

## Current behavior

`limactl create -y --name <id>` uses default template with `~` mounted.

## Proposed fix

Generate a custom Lima template per agent instance:

```yaml
mounts:
  - location: "~/.carrier/instances/<agent-id>"
    writable: true
  # No home directory mount

provision:
  - mode: system
    script: |
      apt-get update -qq && apt-get install -y -qq bubblewrap
```

### Changes needed

1. Generate Lima YAML template in `limaIsolationBackend.PrepareCommands()` — write to `~/.carrier/lima/<agent-id>.yaml`, use `limactl create -y --name <id> <template-path>`
2. Mount only agent workspace — no `~`
3. Auto-install bwrap in VM provision script
4. Optional network policy flag: `--isolation-network=none|host|restricted`

## Acceptance criteria

- [ ] Custom Lima template generated per agent
- [ ] Only agent workspace directory mounted (writable)
- [ ] No `~` mount by default
- [ ] bwrap auto-provisioned in VM
- [ ] Network isolation flag
- [ ] Tests for template generation

dev01lay2:
Closes #1493

### 1. Custom Lima template — restrict mounts
- Generates `~/.carrier/lima/<instance>.yaml` with only agent workspace mounted (writable)
- **No `~` mount** — agent cannot read host SSH keys, AWS creds, etc.
- Auto-provisions bwrap + git + curl in VM via provision script
- Uses `limactl create -y --name <id> <template>` instead of bare create

### 2. VM cleanup on uninstall
- `carrier uninstall` now calls `limactl stop` + `limactl delete` for isolated agents
- Also cleans up the Lima template file
- Best-effort: errors logged but don't block uninstall
- WSL cleanup path included for Windows

### 3. Track isolation in AgentState
- Added `Isolated bool` field to `AgentState`
- Set on start when `Isolation: true`
- Persisted so uninstall knows whether to clean up VM
- VM cleanup only runs if `state.Isolated == true`

## Testing
All daemon tests pass.

### DIFF
--- a/daemon/internal/lifecycle/isolation_backend.go
+++ b/daemon/internal/lifecycle/isolation_backend.go
@@ -50,7 +50,8 @@ type isolationBackend interface {
 }
 
 type linuxIsolationBackend struct {
-	bwrapPath string
+	bwrapPath    string
+	agentWorkDir string
 }
 
 func (b linuxIsolationBackend) CommandGOOS() string {
@@ -66,7 +67,7 @@ func (b linuxIsolationBackend) WrapCommand(command string) (string, error) {
 }
 
 func (b linuxIsolationBackend) WrapStartCommand(startCommand string) (string, error) {
-	return buildBwrapInvocation(b.bwrapPath, startCommand)
+	return buildBwrapInvocation(b.bwrapPath, startCommand, b.agentWorkDir)
 }
 
 func (b linuxIsolationBackend) PrepareCommands() ([]string, error) {
@@ -221,7 +222,14 @@ func resolveIsolationBackend(agentID, agentWorkDir string) (isolationBackend, er
 		if err != nil || strings.TrimSpace(bwrapPath) == "" {
 			return nil, fmt.Errorf("%w: bubblewrap (bwrap) executable not found in PATH", ErrIsolationUnavailable)
 		}
-		return linuxIsolationBackend{bwrapPath: strings.TrimSpace(bwrapPath)}, nil
+		workDir := strings.TrimSpace(agentWorkDir)
+		if workDir == "" {
+			workDir = strings.TrimSpace(isolationEnvLookup(isolationWorkDirEnvKey))
+		}
+		return linuxIsolationBackend{
+			bwrapPath:    strings.TrimSpace(bwrapPath),
+			agentWorkDir: workDir,
+		}, nil
 	case manifest.CommandOSDarwin:
 		limactlPath, err := isolationBackendLookup("limactl")
 		if err != nil || strings.TrimSpace(limactlPath) == "" {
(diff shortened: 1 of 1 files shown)


daemon/internal/lifecycle/isolation_backend.go
```diff
@@ -184,9 +232,17 @@ func resolveIsolationBackend() (isolationBackend, error) {
 		if instance == "" {
 			instance = defaultLimaInstanceName
 		}
```
Keith-CY: BS: This still reuses global instance `default` for all agents. Isolation isolation no longer guarantees isolation and allows cross-agent access to mounted workspaces. Add per-agent instance IDs and pass the selected `agentID` into `resolveIsolationBackend`/`StartWithOptions`/`InstallWithOptions`.

daemon/internal/lifecycle/isolation_backend.go
```diff
@@ -101,15 +104,60 @@ func (b limaIsolationBackend) WrapStartCommand(startCommand string) (string, err
...
+mounts: []
+provision:
+  - mode: system
+    script: |
+      apt-get update -qq && apt-get install -y -qq bubblewrap git curl
+`) + "\n"
+	} else {
+		templateContent = fmt.Sprintf(strings.TrimSpace(`
+mounts:
+  - location: "%s"
```
Keith-CY: BS: `agentWorkDir` is interpolated into YAML via `fmt.Sprintf` before quoting. If it contains `"` or newlines, this enables template injection and arbitrary mount/provision sections. Encode/sanitize with a safe serializer instead of string interpolation.

daemon/internal/lifecycle/uninstall.go
```diff
@@ -47,3 +58,69 @@ func (s *Service) Uninstall(ctx context.Context, agentID string) error {
...
+	return nil
+}
+
+func cleanupLimaTemplateFile(instance string) error {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("resolve user home directory for Lima template cleanup: %w", err)
+	}
+
+	templatePath := filepath.Join(homeDir, ".carrier", "lima", strings.TrimSpace(instance)+".yaml")
```
Keith-CY: BS: `cleanupLimaTemplateFile` joins untrusted instance names directly into a path and deletes it. A crafted instance ID can include traversal and remove arbitrary host files during uninstall. Validate instance IDs to trusted pattern before path construction.

Akane-CN (changes requested):
I re-reviewed isolation-state tracking and found one blocking state regression.

daemon/internal/lifecycle/isolation_backend.go
dev01lay2: Addressed in 06e457b — `linuxIsolationBackend` now carries `agentWorkDir` and `buildBwrapInvocation` uses it to restrict mounts: `--ro-bind / /` (read-only root) + writable `--bind` only for the agent's workspace. No more shared full-root mount across agents. Guest bwrap inside Lima/WSL keeps full mount since the VM already provides per-instance isolation.

PR #1498, opened 2026-02-28
fix(memory): port and harden FusionMem V2 on main

Keith-CY:
## Summary
- port FusionMem V2 memory stack to origin/main (truth files + sqlite index + migration APIs)
- preserve per-agent/shared/public + instance attach/detach/import/export compatibility
- harden post-review issues:
  - avoid read-lock write race in sqlite-backed search path
  - preserve explicit instance scope grants when detaching memory attachments
  - restrict rollback backups to artifacts/migration/*.zip
  - validate migration consistency using active (non-archived) records only
- add regression tests for explicit-scope preservation and rollback path guard

## Validation
- go test ./internal/memory/... (daemon)
- go test ./server/... (daemon)
- go test ./... (baseagent)
- go test ./cmd/carrier -run TestNonExistent

## Notes
- go test ./... in daemon still has platform-specific lifecycle failures on macOS in this environment (collectAgentMetrics only supports Linux, isolation backend expectation tests).

### DIFF
--- a/scripts/coverage-gate.sh
+++ b/scripts/coverage-gate.sh
@@ -9,7 +9,9 @@ STRICT_MODE="${COVERAGE_STRICT_100:-0}"
 
 threshold_shared="${COVERAGE_THRESHOLD_SHARED:-100.0}"
 threshold_baseagent="${COVERAGE_THRESHOLD_BASEAGENT:-83.0}"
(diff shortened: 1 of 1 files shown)


dev01lay2 (approved):
FusionMem V2 port to main with good hardening: rollback path restricted to artifacts/migration, explicit scope preservation on detach, read-lock safety in SQLite search. Well-tested. LGTM.

PR #1501, opened 2026-02-28
fix: resolve CI flaky failures (crypto/rand, TempDir, Playwright)

dev01lay2:
### 1. Gateway tests: crypto/rand exhaustion
- **Symptom**: `fatal error: crypto/rand: failed to read random data`
- **Cause**: Go 1.24 + `-race` with many concurrent goroutines exhausts `/dev/urandom` on CI runners
- **Fix**: Add `-parallel=4` to gateway test command to limit concurrent test goroutines
- **Ref**: https://go.dev/issue/66821

### 2. Isolation e2e tests: TempDir cleanup failure
- **Symptom**: `TempDir RemoveAll cleanup: permission denied` on Go module cache files
- **Cause**: `go build` inside `t.TempDir()` creates read-only module cache files; Go's cleanup can't delete them
- **Fix**: Add `t.Cleanup(func() { makeTreeWritable(tmp) })` before TempDir cleanup runs

### 3. WebUI auth-expiry: Playwright timeout
- **Symptom**: `auth-expiry.spec.ts` tests fail intermittently waiting for `#login-overlay`
- **Cause**: Default 5s timeout too tight for 401→redirect→overlay render chain on slow CI
- **Fix**: Increase `toBeVisible()` timeout to 10s for the three auth-expiry assertions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,7 +220,7 @@ jobs:
 
       - name: Gateway tests
         working-directory: gateway
-        run: go test -race -count=1 ./...
+        run: go test -race -count=1 -parallel=2 -timeout=5m ./...
 
   coverage-gate:
     name: Coverage Gate
--- a/cmd/carrier/main_e2e_test.go
+++ b/cmd/carrier/main_e2e_test.go
@@ -20,6 +20,24 @@ import (
 	"carrier/configv2"
 )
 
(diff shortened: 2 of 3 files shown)


Akane-CN (changes requested):
I reviewed the latest revision. CI is green, but there is one blocking quality regression in the WebUI auth-expiry coverage change.

PR #1530, opened 2026-03-01
Enhance memory search options and hybrid reranking

Keith-CY:
## Summary
- add new memory search options (candidateMultiplier, adaptiveRecall, rerank, lexicalWeight, semanticWeight)
- implement hybrid lexical+semantic reranking and candidate tuning in Fusion search
- add in-memory fallback and deterministic search weighting helpers
- extend /api/v2/memory/search request parsing to support new options
- add regression coverage for reranking behavior

## Testing
Not run in this request (per instructions).

### DIFF
--- a/daemon/internal/memory/fusion_store.go
+++ b/daemon/internal/memory/fusion_store.go
@@ -24,7 +24,7 @@ const (
 	maxCandidateLimit       = 300
 	minCandidateLimit       = 20
 	defaultLexicalWeight    = 0.65
-	defaultSemanticWeight   = 0.35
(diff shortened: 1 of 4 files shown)


Keith-CY:
Implemented both points from Gemini review:

1. Semantic scoring now relies only on token-set matching (no substring fallback), which avoids false-positive matches like query `or` matching `for`.
2. Removed the redundant `ratio > 1` clamp in `scoreText` because match count is already bounded by query token count.

PR now includes commit `3c2cb41` on branch `codex/memory-search-options`.

PR #1538, opened 2026-03-02
test: raise module coverage above 95%

Keith-CY:
## Summary
- Add focused test coverage updates for gateway modules without changing non-test code.
- Keep behavior untouched; only new/extended test cases are added.

## Testing
- New tests added across gateway modules, including pairing sessions, remote keys, feature flags, remote audit, managed instances, telegram transport status, and credential stores.
- PR intended for continuing with more coverage updates in follow-up commits if needed.

### DIFF
--- a/gateway/remote_keys_api_more_test.go
+++ b/gateway/remote_keys_api_more_test.go
@@ -11,6 +11,16 @@ import (
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


dev01lay2 (approved):
Re-reviewed after new push. The incremental change adds a `getRemoteKeyErrorMessage` helper to handle both `errorMessage` and `message` response fields — clean fix. LGTM.

PR #1561, opened 2026-03-12
feat: advance baseagent parity and gateway auth flows

Keith-CY:
## Summary
- add baseagent web/delegation tools, skills, managed MCP, and Carrier memory integration
- unify gateway channel registry, inbound routing, pairing/session auth, and provider/channel auth status APIs
- align WebUI onboarding/settings with the unified channel/auth model and harden integration coverage

## Testing
- go test ./... in gateway
- go test ./server in daemon
- go test ./... in baseagent
- bun run test in webui

### DIFF
--- a/gateway/orchestrator_api_more_test.go
+++ b/gateway/orchestrator_api_more_test.go
@@ -2667,17 +2667,15 @@ func TestRunOrchestratorTaskAttemptGetHostErrorBranch(t *testing.T) {
 
 func mustMakeReadOnly(t *testing.T, path string) {
 	t.Helper()
-	info, err := os.Stat(path)
-	if err != nil {
-		t.Fatalf("stat path %q failed: %v", path, err)
-	}
-	if info.IsDir() {
-		if err := os.Chmod(path, 0o500); err != nil {
-			t.Fatalf("chmod dir read-only failed: %v", err)
+	target := strings.TrimSpace(path)
+	prev := writeRemoteControlStoreFile
+	writeRemoteControlStoreFile = func(name string, data []byte, perm os.FileMode) error {
+		if strings.TrimSpace(name) == target {
+			return os.ErrPermission
 		}
-		return
+		return prev(name, data, perm)
 	}
-	if err := os.Chmod(path, 0o400); err != nil {
-		t.Fatalf("chmod file read-only failed: %v", err)
-	}
+	t.Cleanup(func() {
+		writeRemoteControlStoreFile = prev
+	})
 }
--- a/gateway/remote_store.go
+++ b/gateway/remote_store.go
@@ -33,6 +33,7 @@ type providerProfilePatch struct {
 }
 
 var remoteControlStoreMu sync.Mutex
+var writeRemoteControlStoreFile = os.WriteFile
 
 func remoteControlStorePath() (string, error) {
 	if custom := strings.TrimSpace(os.Getenv("CARRIER_REMOTE_CONTROL_STORE")); custom != "" {
@@ -111,7 +112,7 @@ func saveRemoteControlState(path string, state *remoteControlState) error {
 	if err != nil {
 		return fmt.Errorf("marshal remote control store: %w", err)
 	}
-	if err := os.WriteFile(path, append(raw, '\n'), 0o600); err != nil {
+	if err := writeRemoteControlStoreFile(path, append(raw, '\n'), 0o600); err != nil {
 		return fmt.Errorf("write remote control store: %w", err)
 	}
 	return nil


Akane-CN (commented):
This is a strong step forward on the baseagent/gateway parity work, but I’m not comfortable approving it yet because I hit concrete local regressions in the touched `gateway` package.

What I validated locally:
- `cd baseagent && go test ./...` ✅
- `cd gateway && go test ./...` ❌

The `gateway` suite is failing in error-path tests that now return success / partial-success statuses instead of the previously expected persistence-error responses. The failures I hit locally include:

- `TestHandleOrchestratorExecutionsErrorBranches/create-save-error`
  - expected `500`, got `201`
- `TestHandleOrchestratorExecutionsErrorBranches/authorize-decline-save-error`
  - expected `500`, got `200`
- `TestHandleOrchestratorExecutionsErrorBranches/authorize-approve-save-error-and-concurrency-defaults`
  - expected `500`, got `202`
- `TestProvisionOrchestratorWorkersErrorBranches/second-lease-save-error-after-config-check`
- `TestHandleWebUIInstance_ErrorBranches/start_persistence_error`
  - expected `500`, got `200`
- `TestHandleWebUIInstance_ErrorBranches/stop_persistence_error`
  - expected `500`, got `200`
- `TestHandleWebUIInstance_ErrorBranches/uninstall_persistence_error`
  - expected `500`, got `200`

So at minimum, this PR is changing persistence/error semantics in `gateway` in ways that the current test suite does not accept. That may be intentional, but if so I’d want the contract change called out explicitly and the affected tests updated in the same PR. Right now it looks like an unintended behavioral drift from nearby refactors.

I didn’t treat the `webui` run as a blocker because that local Bun/Vitest environment is noisy here, but the `gateway` failures above are specific and reproducible.

Happy to re-review after the gateway persistence/error-path behavior is reconciled.

Keith-CY:
Re-checked this on current head `96635591` because the reported failures are all in touched `gateway` persistence/error-path coverage.

What I ran locally in this worktree:

- `cd gateway && go test -count=1 ./...`
- `cd gateway && go test -count=1 -run 'TestHandleOrchestratorExecutionsErrorBranches|TestProvisionOrchestratorWorkersErrorBranches|TestHandleWebUIInstance_ErrorBranches' ./...`

Both pass on `96635591`.

So I can’t reproduce the reported status drift on the current PR head. If you still see those cases returning `201` / `202` / `200` instead of `500`, please paste the failing test output and the exact commit/worktree state you ran against. That would let me distinguish between a real current-head regression and a stale local/base mismatch.

Keith-CY:
Follow-up on Akane's reproducible `gateway` persistence-error report:

I checked the exact three `TestHandleWebUIInstance_ErrorBranches` cases again and the handler path is still returning `E_STATE_PERSISTENCE` correctly here. The unstable part turned out to be the test setup, not the action handler logic.

Those cases were simulating persistence failure via `chmod 0400` on the instance store file. That can vary across runtime user / filesystem setups. I pushed `4001ea58`, which makes the failure deterministic by replacing the store file with a directory after the initial load and before the save path executes. That guarantees `saveManagedInstances` fails in the action handler without depending on file-mode semantics.

Re-verified on `4001ea58`:
- `cd gateway && go test -count=1 -run 'TestHandleWebUIInstance_ErrorBranches/(start_persistence_error|stop_persistence_error|uninstall_persistence_error)$' ./...`
- `cd gateway && go test -count=1 ./...`

Both pass.

Keith-CY:
Follow-up on Akane's orchestrator persistence-error report:

I checked the four current-head failures you listed and the underlying issue was the same class as the WebUI branch: the tests were relying on filesystem permission bits to simulate a write failure.

I pushed `439cd54e`, which makes those orchestrator persistence/error-path cases deterministic by routing `saveRemoteControlState` through a tiny package-level write seam and having the test helper force `os.ErrPermission` for the target store path. That keeps production semantics unchanged while removing the runtime-user/filesystem dependency from the test setup.

Re-verified on `439cd54e`:
- `cd gateway && go test -count=1 -v -run 'TestHandleOrchestratorExecutionsErrorBranches/(create-save-error|authorize-decline-save-error|authorize-approve-save-error-and-concurrency-defaults)$|TestProvisionOrchestratorWorkersErrorBranches/second-lease-save-error-after-config-check$' ./...`
- `cd gateway && go test -count=1 ./...`

Both pass.

Akane-CN (approved):
Re-reviewed on the latest head from a clean checkout.

The previously blocking `gateway` persistence/error-path regressions are resolved on my side now. I re-ran:

- `cd gateway && go test -count=1 -run 'TestHandleOrchestratorExecutionsErrorBranches|TestProvisionOrchestratorWorkersErrorBranches|TestHandleWebUIInstance_ErrorBranches' ./...`
- `cd gateway && go test -count=1 ./...`

Those pass locally on the current PR head.

Given that, I’m comfortable approving this now. Nice work pushing through the parity + auth/channel unification cleanup.

PR #1603, opened 2026-03-23
Add delegation context brokerage and disable OpenRouter live smoke

Keith-CY:
## Summary
- add delegation contracts and blocking subagent context brokerage with daemon/gateway APIs and focused coverage
- extend subagent jobs to track context requests, responses, degraded execution, and outcomes
- disable OpenRouter live-provider smoke jobs in CI while keeping OpenAI live-provider checks enabled

## Test Plan
- go test carrier/baseagent carrier/daemon/server carrier/gateway


baseagent/subagent_manager.go
```diff
@@ -235,6 +317,125 @@ func (m *InMemorySubagentManager) Cancel(_ context.Context, jobID string) error
...
+		}
+		job.MissingContext = trimStringList(append(job.MissingContext, missing))
+		job.Status = SubagentJobStatusDegraded
+		job.Confidence = "low"
+	} else {
+		job.Status = SubagentJobStatusRunning
+		if strings.TrimSpace(job.Confidence) == "" {
+			job.Confidence = "medium"
+		}
+	}
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 52c1de3c. `finalizeContextResponse` now only restores `running` when the job is still `awaiting_context`, so a late fulfilled response can no longer overwrite a degraded/completed terminal state. Added regression coverage in `TestSubagentManagerLateFulfilledResponseDoesNotResetCompletedDegradedJob`.

baseagent/subagent_manager.go
```diff
@@ -204,11 +237,60 @@ func (m *InMemorySubagentManager) RecentJobs(_ context.Context, limit int) []Sub
...
+
+	if hasWaiter {
+		select {
+		case waiter <- normalized:
+		default:
+		}
+		return normalized, nil
+	}
+
+	m.finalizeContextResponse(jobID, requestID, normalized)
```
Akane-CN, line 302: BS: this fallback currently accepts any `requestID` as long as the job exists. If the caller invents an ID (or a stale client responds after the waiter is gone), we still persist a synthetic response and return success. Please verify that `requestID` belongs to `job.ContextRequests` before calling `finalizeContextResponse`; late responses for known requests can stay allowed, but unknown IDs should return not found.

baseagent/subagent_context_broker.go
```diff
@@ -0,0 +1,135 @@
...
+	if response.Status != DelegationContextStatusFulfilled {
+		missing := requestID
+		for _, item := range job.ContextRequests {
+			if strings.TrimSpace(item.RequestID) == requestID && strings.TrimSpace(item.Question) != "" {
+				missing = strings.TrimSpace(item.Question)
+				break
+			}
+		}
+		job.MissingContext = trimStringList(append(job.MissingContext, missing))
+		job.Status = SubagentJobStatusDegraded
```
Akane-CN: BS: this can downgrade a terminal job back to `degraded`. A concrete case is: the request times out, the executor finishes, the job becomes `completed`, and then a late `partial`/`denied` response arrives here. Flipping the status after completion breaks terminal-state assumptions (for example history pruning). Please only mutate `job.Status` while the job is still in-flight (`awaiting_context` / `running`) and keep late responses as audit data on completed jobs.

Akane-CN (commented):
CI is still red right now (coverage gate / metrics harness / darwin app packaging), so I'm not leaving a final approve/request-changes decision yet. I did leave two inline BS notes that look blocking to me:

- reject responses for unknown context-request IDs instead of persisting synthetic responses
- keep terminal jobs terminal when late non-fulfilled responses arrive

Suggested follow-up: add regression tests for both the fabricated-request-id path and the late partial/denied response path.

baseagent/subagent_manager.go
Keith-CY: Fixed in a3ffe4ac. `RespondContextRequest` now verifies that `requestID` belongs to `job.ContextRequests` before accepting either the waiter path or the fallback finalize path, and returns `not found` for fabricated/stale IDs. Added regression coverage for the unknown-request-id case in `TestSubagentManagerRespondContextRequestRejectsUnknownRequestID`.

baseagent/subagent_context_broker.go
Keith-CY: Fixed in a3ffe4ac. `finalizeContextResponse` now treats terminal jobs as audit-only for late responses: it still records the response, but it no longer mutates `Status`, `MissingContext`, or `Confidence` once the job is already terminal. Added regression coverage in `TestSubagentManagerLatePartialResponseDoesNotResetCompletedJob`.

Akane-CN:
Reviewed the latest update (`a3ffe4a`) first.

The two earlier blocking points look addressed on code review:
- unknown context-request IDs are now rejected instead of being persisted
- late non-fulfilled responses no longer downgrade completed jobs

I also re-ran the targeted Go tests locally (`carrier/baseagent`, `carrier/daemon/server`, `carrier/gateway`) and they passed.

I don't see a new blocker in this delta, but I'm still holding the final review decision until CI finishes.

Linn-San (approved):
Reviewed and approved with No BS. I spot-checked the delegation context brokerage flow end-to-end: new delegation contract/context request-response types, subagent manager state transitions (`awaiting_context` / `degraded`), daemon+gateway APIs for listing/responding to context requests, and the CI change that disables OpenRouter live-provider jobs while preserving explicit skip reporting. The tests cover the critical degraded/late-response/cancel paths, and I did not find blocking issues in scope.

Akane-CN (approved):
Re-reviewed the latest code after the follow-up fixes.

The earlier BS findings I raised look addressed now:
- unknown context-request IDs are rejected
- late non-fulfilled responses no longer rewrite completed job state

CI is green, and I also re-ran the targeted Go tests locally (`carrier/baseagent`, `carrier/daemon/server`, `carrier/gateway`). Approving.

Keith-CY:
CI is green now, and both blocking review points are addressed in `a3ffe4ac`:

- unknown `requestID` responses are now rejected
- late non-fulfilled responses no longer change terminal job status

I replied on both inline threads with the implementation details and regression tests. Please re-check when you have a moment.

